PR #13, opened 2016-05-05
Support for AF_VSOCK virtio backend

ijc25:
This is based on [v6 of the proposed virtio-sock protocol](http://thread.gmane.org/gmane.comp.emulators.virtio.devel/1122) and developed against [v5 of the corresponding Linux frontend patches](http://thread.gmane.org/gmane.linux.kernel.virtualization/27455). This PR should not be merged until at least the spec is finalised and the implementation updated to conform.

When creating the device several arguments are required:
* `guest_cid=`: The CID to use for the guest (>=3)
* `path=`: The path to a directory where the backend Unix sockets will live

There is one optional argument:
* `guest_forwards=`: a semi-colon separate list of port numbers to create forwards for

The backend proxies vsock connections into Unix domain sockets created in the configured `path`.

To open a listening socket on the host create a listening Unix domain socket at `sprintf("%s/%08"PRIx32".%08"PRIx32", path, cid, port)` (`cid` is normally `2` for the host).

To connect to a listening vsock from the host open the Unix domain socket at `path/connect` and write to it a string of (precisely) the form `sprintf("%08"PRIx32".%08"PRIx32"\n", cid, port)`. The socket will then be connected.

Alternatively if `guest_forwards` are configured then the hypervisor will be listening on a Unix socket of the form `sprintf("%s/%08"PRIx32".%08"PRIx32", path, guest_cid, port)` which is directly connected to the guest port (no need to write the magic string). A symlink `sprintf("%s/guest.%08"PRIx32", path, port)` is also created for convenience.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,70 @@
+# Contributing to Docker open source projects
+
+Want to hack on this project? Awesome! Here are instructions to get you started.
+
+This project is a part of the [Docker](https://www.docker.com) project, and follows
+the same rules and principles. If you're already familiar with the way
+Docker does things, you'll feel right at home.
+
+Otherwise, go read Docker's
+[contributions guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md),
+[issue triaging](https://github.com/docker/docker/blob/master/project/ISSUE-TRIAGE.md),
+[review process](https://github.com/docker/docker/blob/master/project/REVIEWING.md) and
+[branches and tags](https://github.com/docker/docker/blob/master/project/BRANCHES-AND-TAGS.md).
+
+For an in-depth description of our contribution process, visit the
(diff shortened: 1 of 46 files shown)


ijc25:
The Linux frontend support was merged into Linux mainline with commit [0803e04011c2](http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=0803e04011c2e107b9611660301edde94d7010cc) in time for v4.8-rc1.

So I think the time has come to merge the backend into hyperkit master.

ijc25:
The code currently corresponds to V7 of the spec which was posted to the virtio-spec list in \<1470324277-19300-1-git-send-email-stefanha*@*redhat.com\>.

ijc25:
Older revisions of the docker 4 mac build system expect this branch to exist, so do not remove until the current set of PRs has been flushed.

PR #23, opened 2016-05-18
Improve documentation for tinycore setup.

pnasrat:
Tested commands by updating tinycore to 7.x on OS X and running using
hyperkitrun.sh


test/tinycore.txt
```diff
@@ -2,8 +2,15 @@ These are binaries from
...
+curl -O http://www.tinycorelinux.net/6.x/x86/release/distribution_files/core.gz
+mv core.gz initrd.gz
 mkdir initrd
-( cd initrd ; zcat ../initrd.gz | sudo cpio -idm )
-sudo sed -i '/^# ttyS0$/s#^..##' initrd/etc/securetty 
-sudo sed -i '/^tty1:/s#tty1#ttyS0#g' initrd/etc/inittab
+( cd initrd ; gzcat ../initrd.gz | sudo cpio -idm )
+sudo sed -i -e '/^# ttyS0$/s#^..##' initrd/etc/securetty 
+sudo sed -i -e '/^tty1:/s#tty1#ttyS0#g' initrd/etc/inittab
 ( cd initrd ; find | sudo cpio -o -H newc ) | gzip -c > initrd.gz
```
ijc25: Did this final `find` (in the context not the modified code) really work for you on OSX? I think it needs to be `find .`. What I see locally is:
```
$ ( cd initrd ; find | sudo cpio -o -H newc ) | gzip -c > initrd.gz
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
1 block
```

Perhaps it would be best to pull this stuff out into an actual shell script rather than embedding it in a txt file?  (actually, really a renaming in practice I think since there isn't much else in the txt)

Lastly if we are going to include a checksum IMHO it should be something more secure than md5, like  `shasum` (AKA sha1) or `shasum -s 256` and we should include it for both the kernel and the initrd. But it seems like upstream only publishes the md5 (and unsigned), so probably we should just omit it rather than providing a false sense of security.

pnasrat: Yeah sorry I missed updating that - I used `find .` (if you do `../` you'll have initrd/ in the path so the boot will probably not find init

The intent for the md5sum was just to sanity check the vmlinuz I linked to was in fact the vmlinuz in git

I can easily refactor this PR into a script for both this and for userboot.so for FreeBSD. I want to potentially look at other OS support or testing boots so this might make sense.

ijc25: Currently our release testing for hyperkit is what is in `circle.yml` (i.e. not very much). In the context of Docker for Mac it of course gets a whole bunch more but that is all specific to that environment and doesn't use the stuff here in `test` at all (that stuff is all inherited from xhyve).

I'm very much in two minds about committing the binaries vs downloading at test time. The images are ~10M which isn't too much overall to either download or checkin, but in both cases over time it will add up to a lot of either network traffic or git repo size (which == network traffic for the clone).

Downloading is probably my slight preference, it does add a dependency on an external resource, but it is also the easier decision to change our minds about since it's not as permanent as committing blobs to the repo.

So, I think I've just about talked myself into saying we should download ephemerally ;-).

BTW I'd like eventually to switch to using Alpine as our default Linux test case, since that more closely matches our downstream usage. Not something for you to worry about though.

pnasrat: Thanks. I'll send an updated PR later. Updating to Alpine isn't hard but I
can do that in a follow up.

pnasrat:
FYI tinycorelinux.net seems to be down so I might as well do the alpine move and open a new Pull Request.

PR #25, opened 2016-05-19
vsock latency improvements

ijc25:
These changes take the latency of a simple ping test (small request sent from the host Unix domain socket to a vsock echo server in the guest, 500,000 times) from having a 50<sup>th</sup> percentile of 80-90µs and a 90<sup>th</sup> of ~110-140µs (with a lot of jitter across different runs) to a 50<sup>th</sup> of 40-50µs and a 90<sup>th</sup> of ~75µs (and with considerably less jitter). The overall throughput for large request remains around 140-150MiB/s, although there is an initial drop (to ~120MiB/s) with the initial switch to using the lists this is regained by the later changes (primarily amortizing the rx kick it seems).

### DIFF
--- a/src/pci_virtio_sock.c
+++ b/src/pci_virtio_sock.c
@@ -1660,37 +1660,53 @@ static void *pci_vtsock_rx_thread(void *vsc)
 		maxfd = sc->rx_wake_fd;
 		nrfd = 1;
 
-		if (poll_socks) {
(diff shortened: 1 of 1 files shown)


djs55:
Good results -- LGTM!

rneugeba:
I'm not really familiar with the code, but the changes LGTM. The simplification of list handling and locking seem to make sense, so does the removal of GTOD.

PR #28, opened 2016-05-23
vsock: Add socks which require a credit update to the rx_queue

ijc25:
even if they don't need to be polled. Otherwise a socket which e.g. has
"local_shutdown & SHUTDOWN_TX" won't ever send a credit update and so guest TX
will stall.
    
Plus a fixup to debug mode.


rneugeba:
I'm not very familiar with the vsock code, but the changes seem to match your description. And the description seems to explain the issue we were seeing. So LGTM

PR #29, opened 2016-05-24
vsock: don't block rx in select if there are pending credit updates

ijc25:
Commit a966247ad850 ("vsock: Add socks which require a credit update to the
rx_queue") ensured that sockets which required a credit update were included in
the rx queue even if they had no rx activity to process, however since they are
not included in the set of select()d fds the thread can block waiting for
activity on one of the other threads rather than processing the pending update.

If there are such pending updates then pass a zero timeout to select, this will
poll the status of all the active fds (so we will process what work exists),
but not block so we will immediately process the credit updates too.

If we are not polling sockets at all (poll_socks = false) then we are waiting
for more descriptors and couldn't send a credit update, so don't do anything
with socks requiring a credit update in that case.

### DIFF
--- a/src/pci_virtio_sock.c
+++ b/src/pci_virtio_sock.c
@@ -1639,6 +1639,11 @@ rx_done:
 	while (1) {
 		int nrfd, maxfd, nr;
 		bool did_some_work = true;
+		bool pending_credit_updates = false;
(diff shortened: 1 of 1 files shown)


rneugeba:
Description and patch make sense. LGTM

PR #30, opened 2016-05-24
Name each pthread as an aid to debugging.

ijc25:
Only appears to be supported by lldb and "Saved crash reports" today but not
gdb or instruments.


rneugeba:
LGTM

PR #31, opened 2016-05-24
Name vsock threads.

ijc25:
Follows on from "Name each pthread as an aid to debugging." in master (PR #30).


src/pci_virtio_sock.c
```diff
@@ -1332,6 +1332,8 @@ static void *pci_vtsock_tx_thread(void *vsc)
 	fd_set rfd, wfd;
 	LIST_HEAD(tx_queue, pci_vtsock_sock) queue;
 
+	pthread_setname_np("vsock-tx");
```
dsheets: The convention in the other PR uses `:` but here `-` is used.

ijc25: Written on different days ;-). I'll change this one to match, thanks.

rneugeba:
LGTM

PR #32, opened 2016-06-01
fix typo

shinriyo:
I fix comment


ijc25:
Thanks for this.

I'm not convinced it is worth deviating from FreeBSD (where this particular file originates) over such a trivial typo. Of course if it were to be fixed upstream we could just update.

It's slightly moot given the above comment but I think a S-o-b is still needed even for trivial changes such as this.

PR #33, opened 2016-06-01
vsock: add a generation counter to source port allocation

ijc25:
Currently the source port for a host initiated connection is "allocated" using
the offset of the struct pci_vtsock_sock datastructure in the socks array (plus
a fixed offset). This opens up the posibility of a source port being reused in
fairly short order, which in turn can mean that an OP_SHUTDOWN sent by the peer
as part of a previous connection can disrupt a new connection.

This is fixed in v6 of the protocol proposal, by requiring an OP_RST as a
response to any OP_SHUTDOWN which results in connection teardown however the
currently posted frontend RFC patches do not yet implement this.

To mitigate this issue until the frontend patches are updated add a 16-bit
per-struct pci_vtsock_sock generation counter and include this in the lower 16
bits of the assigned port number moving the array offset (+1 to avoid assigning
port 0) to the top 16 bits. An old/delayed OP_SHUTDOWN now needs to be 2^16
recycles of a given struct behind to cause the problem described above.

For reference:
* [v5 spec proposal](http://thread.gmane.org/gmane.comp.emulators.virtio.devel/1092)
* [v6 spec proposal](http://thread.gmane.org/gmane.comp.emulators.virtio.devel/1122)
* [discussion of shutdown protocol](http://article.gmane.org/gmane.comp.emulators.kvm.devel/150103) (in response to RFC posting of the frontend patches)

### DIFF
--- a/src/pci_virtio_sock.c
+++ b/src/pci_virtio_sock.c
@@ -211,6 +211,7 @@ struct pci_vtsock_sock {
 	 *     -1	Otherwise
 	 */
 	int fd;
+	uint16_t port_generation;
 	/* valid when SOCK_CONNECTED only */
 	uint32_t local_shutdown, peer_shutdown;
 
@@ -1129,7 +1130,8 @@ static void pci_vtsock_proc_tx(struct pci_vtsock_softc *sc,
 
 	case VIRTIO_VSOCK_OP_SHUTDOWN:
 		if (!sock) {
-			PPRINTF(("TX: SHUTDOWN to non-existent sock\n"));
+			DPRINTF(("TX: SHUTDOWN to non-existent sock "PRIcid"."PRIport"\n",
+				 hdr.dst_cid, hdr.dst_port));
 			goto do_rst;
 		}
 		if (sock->state != SOCK_CONNECTED) {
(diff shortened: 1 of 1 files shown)


ijc25:
Failed due to:
```
opam install --yes sexplib=113.00.00 uri qcow-format ocamlfind
[ERROR] sexplib.113.00.00 is not available because your system doesn't comply
        with ocaml-version >= "4.02.1" & ocaml-version < "4.03".

opam install --yes sexplib=113.00.00 uri qcow-format ocamlfind returned exit code 66
```

Which I think is due to OPAM changes, which I think were resolved via b65388ba8a29c4e0eacf797f6300e64d06e7c4a8 in the master branch. I'll merge up master here.

ijc25:
Merging from master fixed ci/circleci. The docker/dco-signed failure is a false +ve due to the GH webui generated merge commits.

rneugeba:
https://github.com/docker/hyperkit/pull/33/commits/39a941dd74e9cdc8431860513ed9d1e405285b0e ie the generation count looks good to me

src/pci_virtio_sock.c
```diff
@@ -1302,10 +1304,12 @@ static void handle_connect_fd(struct pci_vtsock_softc *sc, int accept_fd, uint32
 	sock->peer_addr.cid = cid;
 	sock->peer_addr.port = port;
 	sock->local_addr.cid = VMADDR_CID_HOST;
-	/* Start at 2^16 to be larger than a TCP port
+	/* Start at 2^16 to be larger than a TCP port, add a
+	 * generation counter to reduce port reuse.
 	 * XXX Allocate properly.
          */
-	sock->local_addr.port = 65536U + (uint32_t)(sock - &sc->socks[0]);
+	sock->local_addr.port = ((uint32_t)(sock - &sc->socks[0] + 1) << 16)
```
ijc25: The +1 avoids assigning port 0x00000000. Since it is `array-offset + 1` for the zeroeth-sock it'll start at 0x00010000 and allocate through 0x0001ffff and then wrap back to 0x00010000 again, for the first sock 0x00020000... etc

PR #36, opened 2016-06-08
Introduce simple boot cycle test for hyperkit.

### LINKED ISSUE #24: Improve automated testing
Currently by the looks of it circle just checks the source can compile and build. To support contributor development of greater OS coverage issue #19 it would be nice if there was a simple test cycle that could be run per OS

I'd imagine a shell based test driver that boots the OS - records console output and does some validation would be a simple first step.

I don't want to implement without discussion so please give me feedback if this sounds of use.

pnasrat:
Fixes #24

Adds dependency on expect, wired in through Makefile.

The expect script replicates the command from hyperkitrun.sh to prevent
requiring a prompt whilst in expect. The dependecy chain is handled in
the Makefile.

Tested both success and failure using make test - adjusting the timeout
to 1 will give an example fail that propgates up to the make return.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ TARGET = build/com.docker.hyperkit
 
 all: $(TARGET) | build
 
-.PHONY: clean all
+.PHONY: clean all test
 .SUFFIXES:
 
 -include $(DEP)
@@ -148,3 +148,10 @@ $(TARGET): $(TARGET).sym
 
 clean:
 	@rm -rf build
+	@rm -f test/vmlinuz test/initrd.gz
+
+test/vmlinuz test/initrd.gz:
+	@cd test; ./tinycore.sh
+
+test: $(TARGET) test/vmlinuz test/initrd.gz
+	@./test_linux.exp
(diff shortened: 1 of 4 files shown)


pnasrat:
You can see the successful test run in 

https://circleci.com/gh/docker/hyperkit/81 under the `make test` section

ijc25:
I notice that the `make test` output seems to be missing the kernel output, despite the apparently correct use of `earlyprintk=serial console=ttyS0`, which makes me concerned that something somewhere is not quite wired up right. However, I can't see what might be wrong just by looking at the code change and test logs. My only half-baked theory is that the `"`'s from the invocation are ending up literally in the kernel command line...

The test output It also has extra blank lines interleaved between each line it does print and seems to print some stuff (e.g. the tinylinux ASCII art logo) twice. Maybe these two are just some side effect of how `expect` works (I'm not really familiar with it) and I'm not inclined to block merging over those if you say it's expected (no pun intended).

There's a typo in the commit log "dependecy".

pnasrat:
Let me look into this morning NY time and also write a TESTING.md

pnasrat:
So I think the newlines are due to getting `\r\n` new lines. It is possible we are also getting a screen redraw `^L` code when the kernel message logs causing the banner to be written twice. I think this is working as expected 😈

PR #37, opened 2016-06-09
Minor optimization of tinycore.sh script

dol:
Optimized the tinycore.sh script:
- Rearranged errexit parameter for sh
- Replace```gzcat``` with ```gzip -dc``` for a better [portability](c9s/perlomni.vim#16)
- Create a temporary directory and do the work within this directory. Easy way to clean up

### DIFF
--- a/test/tinycore.sh
+++ b/test/tinycore.sh
@@ -1,4 +1,5 @@
-#!/bin/sh -e
+#!/bin/sh
+set -e
 
 # These are binaries from a mirror of
 #  http://tinycorelinux.net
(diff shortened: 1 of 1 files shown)


test/tinycore.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/sh -e
+#!/usr/bin/env sh
+set -e
```
ijc25: I don't see the point of `/usr/bin/env` here. Standards basically require /bin/sh to exist on any Unix-y system (unlike for other interpreters like bash or python where env might make sense).

ijc25:
Please `s/zcat/gzcat/` in the commit message to accurately reflect the change.

test/tinycore.sh
dol: Agree. It's my habit to define this that way. I'll change it back.
Do you prefer ```#!/bin/sh -e``` over
```bash
#!/bin/sh
set -e
```

ijc25: Thanks. I've no particular preference on how `-e` gets set.

ijc25:
@pnasrat since you authored this script I'd be interested in any feedback/LGTM you might want to add.

dol:
@pnasrat BTW: The distro.ibiblio.org is quite a slow mirror. I used http://ftp.vim.org/os/Linux/distr/tinycorelinux/ for testing purpose. The mirror list can be found at http://wiki.tinycorelinux.net/wiki:mirrors.

pnasrat:
Hey @dol - thanks for the improvements I only added a mirror in as tinycorelinux.net was temporarily unavailable when I was trying to retest this. Feel free to change that - eventually we can probably use some data to find a reasonable mirror for circleci, the mirage team in Cambridge and the US. I'm not sure how often this script is run.

I think we might also want to call the tmpdir removal in a function and use trap to remove on exit - but that could easily be a follow up. It also sucks that it is a sudo call which can prompt for password (though not in circleci) but I don't have a good portable solution than that - maybe fakeroot not sure if it's in BSD ports if these changes get back propagated.

LGTM once you've addressed @ijc25's comments on the commit message

dol:
The updated commit message and some other suggestion from @ijc25 are already in place. Because I pushed with 'git push -f' GitHub killed the other comments from @ijc25, because his comment on a certain line was replaced by a force push.

pnasrat:
LGTM

dol:
@pnasrat To do a cleanup with trap would be a nice extension. I was looking into a portable solution to avoid sudo, but as you already mentioned, is not an easy thing to do.

pnasrat:
Feel free to follow up with that change or add it here depending on @ijc25 preference (I am not a maintainer) - I'm working on testing and some bigger design proposals but happy to review PRs

PR #39, opened 2016-06-22
Support remote debugging of kernel processes.

pnasrat:
Proposal as per

https://docs.docker.com/opensource/workflow/advanced-contributing/

See the file for complete description.

### DIFF
--- a/debug-design-proposal.md
+++ b/debug-design-proposal.md
@@ -1,0 +1,35 @@
+* Why is this change needed or what are the use cases?
+
+   Currently inspecting a hyperkit running kernel requires detailed knowledge
+   of hyperkit or adding additional compilation flags for tracing. Ideally it
+   should be possible to debug a hosted kernel as if using hardware debugger or
+   similar to how qemu/bochs expose a gdbserver.
+
+* What are the requirements this change should meet?
+
+   Provide a mechanism to enable connecting a debugger (lldb, gdb) to the hosted kernel.
+
+* What are some ways to design/implement this feature?
+
+   Both gdb and lldb support using the gdb protocol for remote target debugging.
+
+    * [GDB Remote Protocol](https://sourceware.org/gdb/current/onlinedocs/gdb/Remote-Protocol.html)
+    * [lldb remote](http://lldb.llvm.org/remote.html)
+    * [lldb protocol extensions](http://llvm.org/viewvc/llvm-project/lldb/trunk/docs/lldb-gdb-remote.txt?view=markup)
+
+    Support can be added by exposing a port that acts as a gdbserver allowing both debuggers to connect options are. 
+
+    * TCP/IP `target remote localhost:1234`
+    * Debug serial port for debugging (eg COM2) `target remote com.docker.hyperkit/debugtty`
+    * Support both TCP/IP and serial
+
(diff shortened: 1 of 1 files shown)


pnasrat:
Design proposal cc @ijc25 - I didn't want to do the work without speaking about this and this looks to be the process.

ijc25:
Thanks!

> Supporting gdb and lldb by using the common protocol without extensions. 

Agreed. IIRC the protocol itself is pretty simple and lightweight.

> supporting both TCP/IP and serial

Agreed, although I don't think the debug serial device needs to be exposed as a serial device (e.g. `COMx`) to the guest, it should just be some sort of suitable device node (`pty` presumably) in the host which talks to the hyperkit process with no guest involvement.

> There could be performance impact so the gdb hooks

It's been a looong time since I've looked at the gdb remote protocol, but IIRC most of the functionality is only used/usable while the world is "stopped" (i.e. by a breakpoint, quiescing all the ~~threads~~vcpus) so the performance impact of the individual peek/poke commands etc on a running system is not too worrisome.

Anything which is poking at the system live would need to use the appropriate mechanisms, but those mechanisms should either already exist for internal use anyway so the debugger can piggyback or really ought to be added irrespective of whether they are to be used by a debugger or not. (Maybe this is wishful thinking on my part and hyperkit makes a lot of use of "this can only be accessed from this thread" thinking, I hope that's not too prevalent though!).

> ensure debugging is possible in a signed build.

That would be great if it turns out to be easy (or even a nop) but I wouldn't worry about jumping through too many hoops to make it work, at least not for a first pass, my guess is that people debugging OS kernels aren't likely to be _too_ worried about whether their hypervisor is sandboxed or not ;-).

pnasrat:
Thanks for the feedback - I agree folks using this are likely to be able to work around limitations. I have two questions before I look at implementation:

1. Is hyperkit the right place to push this into or should I be targetting xhyve? I'm not sure as to the relationships of the projects and how much you wish to diverge.
2. Currently there is a `-g gdbport` and implementation`src/dbgport.c` - that seems FreeBSD specific. Does extending that  make sense or would you rather a separate implementation?

```
allow a remote kernel kgdb to be relayed to the guest kernel gdb stub
via a local IPv4 address and this port.
This option will be deprecated in a future version.
```

ijc25:
WRT *#*1 it is really up to you, arguably bhyve would be the right place to start ;-).

For *#*2 AIUI that's just a simple passthrough to an in guest debug stub via a magic I/O port (which it seems *BSD has drivers for, but many other OSes do not have a stub at all and if they did would likely need code adding to support that particular access method).

IMHO forwarding access to an in guest debugger and ability to inspect the guest via a hypervisor debug interface are orthogonal enough that they can and should be kept separate.

ijc25:
Although the workflow document seems to imply merging documents such as the one here into the repo I don't really see the need, the proposal and the conversation around this is adequately addressed by the contents of and comment on this PR IMHO and can be just as easily (if not more easily) referred to in the future etc.

I think there is broad agreement on the shape of things, so I am going to close this but not merge, please treat this as an acceptance of the proposal rather than a rejection and if you are still interested in implementing what is proposed here please go right ahead!

PR #42, opened 2016-07-05
Improve logging/debugging when select() fails

ijc25:
Retry when appropriate and log the failure reason for other failures.

### DIFF
--- a/src/pci_virtio_sock.c
+++ b/src/pci_virtio_sock.c
@@ -340,6 +340,34 @@ static int max_fd(int a, int b)
 		return b;
 }
 
+/*
+ * Returns >= 0 number of fds on success or -1 to indicate caller
+ * should retry. On any failure which cannot be retried logs and exits.
+ */
+static int xselect(const char *ctx,
+		   int nfds, fd_set *readfds, fd_set *writefds,
+		   fd_set *errorfds, struct timeval *timeout)
+{
+	int rc = select(nfds, readfds, writefds, errorfds, timeout);
+	if (rc >= 0) return rc;
+
+	/*
+	 * http://pubs.opengroup.org/onlinepubs/009695399/functions/select.html
+	 * lists EINTR, EBADF and EINVAL. EINTR is recoverable and should be
+	 * retried.
+	 */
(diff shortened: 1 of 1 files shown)


src/pci_virtio_sock.c
```diff
@@ -340,6 +340,34 @@ static int max_fd(int a, int b)
...
+	if (rc >= 0) return rc;
+
+	/*
+	 * http://pubs.opengroup.org/onlinepubs/009695399/functions/select.html
+	 * lists EINTR, EBADF and EINVAL. EINTR is recoverable and should be
+	 * retried.
+	 */
+	if (errno == EINTR) return -1;
+	/*
+	 * OS select(2) man page lists EAGAIN in addition to the above.
```
djs55: s/OS/OSX/ ?

ijc25: Yep, will fix, thanks!

djs55:
LGTM

Out of interest, do you happen to know what `FD_SETSIZE` is on OSX? Is it worth compiling with `-D_DARWIN_UNLIMITED_SELECT`? The OSX manpage has:

>     select() now returns with errno set to EINVAL when nfds is greater than FD_SETSIZE.  Use a smaller value for nfds
>     or compile with -D_DARWIN_UNLIMITED_SELECT.

ijc25:
> LGTM

Thanks!

> Out of ....

`grep -r FD_SETSIZE /usr/include/` throws up a lot of noise but it looks like it is in the vicinity of 1024.

justincormack:
LGTM

justincormack:
I would just use kqueue if you want bigger sets than select allows, it will perform better anyway

ijc25:
> Is it worth compiling with -D_DARWIN_UNLIMITED_SELECT?

To answer @djs55's actual question: Reading `/usr/include/sys/_select.h` implies the only reason it isn't the default is for "UNIX conformance" (which seems like an odd choice, but oh well), but I'm not sure what the other unspoken trade-offs there might be.

I think I would want to make that change on the master branch, since the use of `select` is not specific to vsock. I'll raise another PR in a moment with that change.

PR #48, opened 2016-07-25
Updated OSSpinLock for Sierra

RnbWd:
While trying to build hyperkit on Sierra (beta 3) - it would throw on `OSSpinLock` - and the error messages suggested it should be replaced with `os_unfair_lock`. So, I replaced the OSSpinLock with the `os_unfair` equivalents  - commented out the `OS_SPINLOCK_INIT` methods, and  It passed the test and works. Here's the  [os API](https://developer.apple.com/reference/os) for reference (doesn't say anything). 

Docker is unusable without disabling the `ntpd` `daemon` - `sudo launchctl unload /System/Library/LaunchDaemons/org.ntp.ntpd.plist` - after that it works perfectly. So I thought that the vmnet / xhyve layer wasn't able to communicate with the system clocks. While testing hyperkit, it has this error message:

````
ACPI Error: Could not enable RealTimeClock event ....
ACPI Warning: Could not enable fixed event - RealTimeClock ....
````

It seems like ACPI  is already a known issue - and the test ended with `PASS`. But while researching why OSSpinLock was depreciated, I ran into. a few articles that made me think it should be replaced and may even be contributing to poor network performance on 10.10+. Webkit wrote a [blogpost](https://webkit.org/blog/6161/locking-in-webkit/) about why they replaced all spinlocks and os-provided-mutexes:

> Compared to OS-provided locks like pthread_mutex, WTF::Lock is 64 times smaller and up to 180 times faster. Compared to OS-provided condition variables like pthread_cond, WTF::Condition is 64 times smaller. Using WTF::Lock instead of pthread_mutex means that WebKit is 10% faster on JetStream, 5% faster on Speedometer, and 5% faster on our page loading test.

I mentioned it in [this issue](https://github.com/docker/hyperkit/issues/47) - and it was pretty simple to implement - but this pull request is meant to raise awareness. I don't know if this method is backwards compatibility < 10.12 - or if they're going to patch 10.10+ to implement this - so it shouldn't be pulled into the main branch. I have not been able to confirm  whether or not OSSpinLock is  potentially effecting the entire hypervisor app - but I haven't seen this discussed anywhere in relation to docker - just wanted to bring it someone's attention.

### DIFF
--- a/include/xhyve/support/platform_locks.h
+++ b/include/xhyve/support/platform_locks.h
@@ -26,7 +26,7 @@
 
 #include <Availability.h>
 
-#if __MAC_OS_X_VERSION_MIN_REQUIRED >= __MAC_10_12
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 1012
 #define XHYVE_USE_OSLOCKS 1
 #include <os/lock.h>
 #else
--- a/include/xhyve/support/platform_locks.h
+++ b/include/xhyve/support/platform_locks.h
@@ -1,0 +1,34 @@
+/*-
+ * Copyright (c) 2016 Anil Madhavapeddy <anil@recoil.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include <Availability.h>
+
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= __MAC_10_12
+#define XHYVE_USE_OSLOCKS 1
+#include <os/lock.h>
+#else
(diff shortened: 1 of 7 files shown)


avsm:
Thanks for the patch!  We should restore build compatibility first with 10.12 and ensure correct operation, and then do the performance profiling.  I've got a patch in avsm/hyperkit@a9bdc65549ddabdc15c507341d8549f45f73f313 that checks for presence of the new functionality and uses it if the minimum version of MacOS X compatibility required is 10.12 or higher.

The minimum version capability is activated by setting the `MACOSX_DEPLOYMENT_TARGET` environment variable to `10.10` (which we need to do in Docker for Mac), at which point the build will break with deprecation warnings if compiled on 10.12.  We can fix this either by:

- disabling `-Werror` if MACOS_DEPLOYMENT_TARGET is lower than 10.12 (a big hammer)
- specifically disabling deprecation warnings, or at least not making them error. Need to look up the `clang` flags that allow this level of precision.

Any thoughts on this @ijc25 @djs55?

avsm:
@RnbWd to make this suitable for merging, if you could pull in my patch, remove the whitespace changes from your diff, and create a single commit with a `Signed-off-by:` line, that would progress things.

rneugeba:
-10 on disabling `-Werror` :)

RnbWd:
@avsm there's more files that need to be patched, which I attempted to do based on your patch, but there's one more variable that throws an error if it's included in 10.12 beta. This is the error being thrown:

````
rc/vmm/vmm.c:217:5: error: use of undeclared identifier 'OS_SPINLOCK_INIT'
    vcpu_lock_init(vcpu);
    ^
src/vmm/vmm.c:97:39: note: expanded from macro 'vcpu_lock_init'
#define vcpu_lock_init(v) (v)->lock = OS_SPINLOCK_INIT;
````

I couldn't find an equivalent method in the os API - but I don't know that much about spinlocks so before finalizing the merge (which I can reduce into a single commit) - I want to make sure that the `init` function is handled properly. I commented it out so that it would compile - but it's only used once in each file where it's defined.

RnbWd:
@GordonTheTurtle thanks - I will do that for the next push - but I need help / review to finish to finalized the change

RnbWd:
I think my commit was rebased properly, although my last commit was not the most descriptive - I attempted to write a fix by using `#ifndef XHYVE_USE_OSLOCKS` - and it compiles on my machine locally, but I am not that experienced with C - so. I hope I patched it correctly. The failure now appears to be in located in `platform_locks.h` where `<os/lock.h>` is included

ijc25:
Hi @RnbWd, sorry for the delay responding I was offline for a long weekend. Anyway, thanks for bringing this to our attention.

I'm a little bit wary of using this undocumented new lock type, but I'm rather more wary of splitting the platform testing base into 2 (<10.12 and >=10.12) through the use of ifdef in this proposed fix (as it stands the newer code wouldn't be tested in CI either).

I think a preferable approach would be to switch unconditionally to using a pthread mutex. Hyperkit already extensively uses pthread's and pthread mutexes extensively in other contexts and I'm unsure what the motivation for using a platform-specific spinlock in this particular context is[0].

Would you be OK with reworking to use pthread mutexes? Don't worry if not, I can try and find someone to take a look.

[0] I suspect (just guessing) that it is something inherited from bhyve where spinlock vs mutex can make a difference when the code is running in kernel context, which is not something hyperkit needs to concern itself with since it is entirely in userspace.

avsm:
>  I suspect (just guessing) that it is something inherited from bhyve where spinlock vs mutex can make a difference when the code is running in kernel context, which is not something hyperkit needs to concern itself with since it is entirely in userspace.

I've looked at the locks and can't see any reason why a pthread_mutex wouldn't suffice.  There are actually mutexes right beside the use of the spinlock in the vapic structures, so there's ample precedent...

RnbWd:
@ijc25 I agree, I just learned about  all of this while attempting to compile hyperkit a few days ago, so I probably can't contribute much more at this time (until I  learn more). I just was curious as to why OSSpinLock was deprecated and the `os_unfair_lock` is still `beta` - but after researching OSSpinLock - it sounded pretty bad. I think having a branch for testing 10.12 would be interesting - but I think a compatible replacement for OSSpinLock makes sense.

ijc25:
@RnbWd thanks, I took a stab at the conversion and am therefore closing this issue in favour of #49 which implements the switch to pthread mutexes.

Issue #47, opened 2016-07-22
OSSpinlock is depreciated


RnbWd:
Is this being used in Docker Mac Beta? I'm running macOS dev beta 3 - and somebody on the forums figured out that the `ntpd daemon` was causing the vm to hang (the normal behavior is for it to start  normally, then after using the network for a few minutes - docker becomes unresponsive. Unloading `ntpd` - docker f works perfectly fine. 

`sudo launchctl unload/System/Library/LaunchDaemons/org.ntp.ntpd.plist`

So I came here to investigate *why* this was happening. When I started to compile hyperkit, the build stopped and there was a warning that said something like

> OSSpinLock is depreciated in version of macOs

Google searched it - found this article: http://engineering.postmates.com/Spinlocks-Considered-Harmful-On-iOS/

Apparently the command is considered harmful and causes networking issues for iOS and macOS - so in Sierra it's not even allowed to be compiled and potentially causing the hang - it's probably causing issues in all versions of OSX but in Sierra it's debilitating. Unless you kill`ntpd`. 

So i went into the docs to find what they are replacing OSSpinLock with, and I found an undocumented command in `os` called `os_unfair_lock` https://developer.apple.com/reference/os/os_unfair_lock

So I started rewriting hacking on hyperkit (and parts of xhyve) to replace it. I've never actually attempted to program C (or even Objective C) - but just wanted to let ya'll know, and if I get it to compile locally - I'll submit a pull request.

I assume the new API would probably be the most efficient replacement - although there may be other ways around this that don't use  `unfair_lock` - I assume that Apple intends that to be a replacement? But the docs don't say anything about it.

RnbWd:
I was able to get hyperkit to compile with the new os API  - tests pass - but there's an error in the that indicates something else is also causing the issue with Sierra. I think some fix here could translate to for all Docker for Mac from Yosemite - but I don't know if the new kernel os methods are supported anywhere else - so it's potentially just a modification for Sierra, for now. But I'll email docker security just in case there's a security issue here that I'm not aware of.

cjsiOS:
Can you tell me  'os_unfair_lock' in which header files？

PR #49, opened 2016-07-27
Switch all OSSpinLock uses to pthread_mutex_t

ijc25:
@RnbWd reported in #47 and #48 that OSSpinLock is deprecated in OSX 10.12 (Sierra, beta 3). Rather than switch to a different lock only for 10.12+ (as was done in #48) this PR instead switches everything over to use pthread mutexes which are available on all versions.

I'm not completely wedded to the `xpthread_mutex_foo` wrappers to do error checking (names by analogy to the more commonly used `xmalloc` wrappers), but I think we certainly need to check at least that `pthread_mutex_lock` actually locked something even though the errors are in the "should never happen category".

### DIFF
--- a/include/xhyve/support/misc.h
+++ b/include/xhyve/support/misc.h
@@ -1,8 +1,11 @@
 #pragma once
 
+#include <assert.h>
+#include <pthread.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 
 #define UNUSED __attribute__ ((unused))
 #define CTASSERT(x) _Static_assert ((x), "CTASSERT")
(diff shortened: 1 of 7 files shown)


include/xhyve/support/misc.h
```diff
@@ -65,3 +68,25 @@ static inline void do_cpuid(unsigned ax, unsigned *p) {
 
 /* Used to trigger a self-shutdown */
 extern void push_power_button(void);
+
+/* Error checking pthread mutex operations */
+static inline void xpthread_mutex_init(pthread_mutex_t *mutex)
+{
+	int rc = pthread_mutex_init(mutex, NULL);
+	assert(rc == 0);
```
avsm: Shouldn't this be a dynamic check that called `xhyve_abort`, since there are valid non-zero return codes from `pthread_mutex_init`?  E.g. `ENOMEM`.  Asserts won't fire if disabled.

ijc25: I lazily looked at Linux's `pthread_mutex_init(1)` (which says init only ever returns 0) since it was closer to hand rather than going to the source http://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_init.html for the actual behaviour. I'll update to follow the same pattern as the other two wrappers.

avsm:
LGTM

PR #56, opened 2016-08-22
Fix local shutdown

ijc25:
The `assert` and the` _rx` suffix on the function are wrong, it is perfectly
fine to call from either rx or tx thread.

In principle it would be possible to have an optimised tx-thread-only version
which could skip the CLOSING_TX state and go right to CLOSING_RX, the current
state prior to this commit is a botched/incomplete attempt at such an
optimisation (but the wrong way round and broken).


rneugeba:
I tested this with the tests where it failed before and now the tests pass. LGTM

PR #61, opened 2016-09-21
Open block devices using the `Block.Config` interface

djs55:
The mirage-block-unix has a revamped disk configuration interface where a `Block.Config.t` represents a desired configuration including parameters

- whether using buffered I/O or not
- whether honouring `flush` or not (possibly more options here in future)
- the path to the file

This patch simplifies the mirage block interface by using `Block.Config.of_string` rather than manually parsing a URI here. The device can then be opened with `Block.of_config`. In future if mirage-block-unix gains some extra parameters, this code won't have to be modified (hopefully)

Note this new interface is not yet in a released version, so this PR includes an extra `opam pin` in the `circle.yml`.

This is part of a workaround for [docker/for-mac#668]

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,6 +10,7 @@ test:
     - brew install opam
     - opam init --yes
     - opam pin add qcow-format git://github.com/mirage/ocaml-qcow#master --yes
+    - opam pin add mirage-block-unix git://github.com/mirage/mirage-block-unix#master --yes
     - opam install --yes uri qcow-format ocamlfind
     - make clean
     - eval `opam config env` && make all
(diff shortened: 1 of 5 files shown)


ijc25:
> Note this new interface is not yet in a released version, so this PR includes an extra opam pin in the circle.yml.

Looks like you might have missed that bit out of the commit?

ijc25:
Code changes LGTM (to the extent of my ability to adjudge ocaml).

djs55:
@ijc25 indeed -- I forgot to `git push` that extra patch!

djs55:
@samoht could you do me a favour and have a quick read through the OCaml code in this PR?

djs55:
Thanks!

PR #59, opened 2016-09-13
Update virt-sock backend to v7 spec

ijc25:
V7 spec was posted to the virtio-spec list in <1470324277-19300-1-git-send-email-stefanha@redhat.com>.

This requires a corresponding update to the Linux frontend to the v6 RFC patches posted as <1469716595-13591-1-git-send-email-stefanha@redhat.com>. These have also been merged into Linux v4.8-rc1.

This should hopefully fix docker/hyperkit#50.

Signed-off-by: Ian Campbell <ian.campbell@docker.com>

PR #62, opened 2016-09-30
Flush any bufered data before closing the socket

ijc25:
v7 of the vsock spec improved the semantics of the protocol on shutdown. Updating to those new semantics (in #59) has exposed a latent bug where TX data which was buffered could be lost because we did `shutdown(SHUT_WR)` on the Unix domain socket while there was still buffered data pending. Avoid this by deferring the `shutdown(SHUT_WR)` to the end of `buffer_drain`, using the difference between `sock->local_shutdown` and `sock->peer_shutdown` to know when this should be done.

There was also a crash because we would repeatedly receive `EPIPE` from the `write` in `buffer_drain` (because we were trying to write to a socket which had been closed in that direction) and hit `assert(s->local_shutdown != VIRTIO_VSOCK_FLAG_SHUTDOWN_ALL)` in `shutdown_local_sock`. The `tx_thread` should be ignoring a socket which has been `SHUT_WR`, so add a check on `local_shutdown`.

### DIFF
--- a/src/pci_virtio_sock.c
+++ b/src/pci_virtio_sock.c
@@ -1502,6 +1502,11 @@ static void *pci_vtsock_tx_thread(void *vsc)
 				}
 			}
 
+			if (s->local_shutdown & VIRTIO_VSOCK_FLAG_SHUTDOWN_RX) {
+				put_sock(s);
+				continue;
+			}
+
 			if (sock_is_buffering(s)) {
 				FD_SET(s->fd, &wfd);
 				maxfd = max_fd(s->fd, maxfd);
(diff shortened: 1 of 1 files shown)


djs55:
LGTM

If we wanted to test this would it suffice to `connect`/`accept` a connection, then `write` a message and immediately and `shutdown` from the client end and then verify that the message is received on the server end?

ijc25:
It's only traffic coming out of the VM, but if the VM end sent a large enough batch data and then did `shutdown(SHUT_WR)` on its end then this should provoke this.

The tricky bit is "large enough batch" since the backend will copy directly from the vring into the Unix domain socket and will only start buffering if it gets a short write indicating that the socket's internal buffers are full. So you'd need to arrange for the server to be slow (or stationary) at reading for some time or something.

In practice `curl`ing a 10M from a webbrowser in a container (in Docker for Mac) was enough to trigger this about 1/3 of the time.

PR #64, opened 2016-10-11
Remove mirage-block-unix pin

djs55:
- remove the pin for `mirage-block-unix` (not needed any more and actually harmful as the master branch has incompatible interfaces)
- add a note to the README to explain that an existing pin to `mirage-block-unix` should be removed
- removed an out-of-date note in the README about the `sexplib` library

/cc @yomimono @ijc25 @rneugeba

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,18 @@ using that to install the appropriate libraries:
     $ brew install opam
     $ opam init
     $ eval `opam config env`
-    $ opam pin add qcow-format git://github.com/mirage/ocaml-qcow#master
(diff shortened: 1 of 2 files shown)


ijc25:
LGTM. I take the it explicit pin of qcow is still needed?

djs55:
Oh actually I don't think the pin is needed there either-- everything needed should be in the latest qcow release. I'll add a patch to remove that pin too.

rneugeba:
LGTM and also fixes my local build :)
merging

PR #65, opened 2016-10-11
[RFC] restructure the layout of the source code

rneugeba:
- All source code moved to ./src (including include files)
- Move bulk of the implementation into a ./lib directory
- Move all test files under ./test
- Remove some un-used files
- Other minor tweaks like

/ping @ijc @avsm


ijc25:
I don't really expect this sort of thing to be reviewable but I had a poke through:
```
git fetch origin
git fetch origin pull/65/head
git diff -C -M origin/master FETCH_HEAD
git diff --stat -C -M origin/master FETCH_HEAD
gitk origin/master..FETCH_HEAD
```

Looks like [src: move the bulk of the source code into ./src/lib](82af13a683a35a94d98945e0422b73a0539bc740) also stripped some trailing whitespace, hopefully no actual semantic changes mixed in, but I wasn't about to actually read the diff to `acpitbl.c` ;-). Likewise [src: move include files into ./src](25edc82298a520df18f46c87b231b362ccc633f5).

FWIW I tend to use `build/opam` as my local private opam, but I gather opam2 will pickup `.opam` automatically, so that's clearly a better choice in the long run.

Is `dsdt.asl` (secretly, and manually) the source to a bunch of the hard coded tables in `acpitbl.c`? I'm not really sure.

Anyway, nothing blocking here, so LGTM

rneugeba:
ah, my git config has:
```
[apply]
    whitespace = fix
```
and must have fixed them up. I just reviewed the changes and they all look fine.

As for `dsdt.asl` I couldn't find any reference in the source. `acpitbl.c` was introduced with [this patch](https://github.com/mist64/xhyve/commit/0e9d07db34b2d44618cf1e0f273ef199dbe82030). The commit suggests that with bhyve the DSDT is generated on  the fly on VM start (using `iasl`) and the patch introduced minimal mostly hard-coded tables. So we should be fine.

rneugeba:
rebased onto #66

PR #66, opened 2016-10-13
kexec: fix wrong calculation method of ramdisk_start

zchee:
Fix `ramdisk_start` promlem in the [kexec.c](https://github.com/docker/hyperkit/blob/master/src/firmware/kexec.c).

- Use the `initrd_addr_max` value from the kernel header if the kernel version is 2.03 or newer.
 - ref: https://github.com/zchee/linux-documentation/blob/master/x86/boot.md#initrd_addr_max
 - Use hardcoded(`0x37ffffff`) value if older.
- Compare the `initrd_max` and `memory.size`, and re-assign if `initrd_max` is larger.
- Add `ALIGNDOWN` macro for `ramdisk_start`, uses prevent to go past the address limit.
- Fix typo for `initrd_addr_max`. thanks @dlorenc

Based on [qemu/hw/i386/pc.c](https://github.com/qemu/qemu/blob/master/hw/i386/pc.c#L954-L979).

Close https://github.com/docker/hyperkit/issues/63

### DIFF
--- a/include/xhyve/firmware/kexec.h
+++ b/include/xhyve/firmware/kexec.h
@@ -29,7 +29,7 @@ struct setup_header {
 	uint8_t ext_loader_ver; /* Extended boot loader version */
 	uint8_t ext_loader_type; /* Extended boot loader ID */
 	uint32_t cmd_line_ptr; /* 32-bit pointer to the kernel command line */
-	uint32_t nitrd_addr_max; /* Highest legal initrd address */
+	uint32_t initrd_addr_max; /* Highest legal initrd address */
 	uint32_t kernel_alignment; /* Physical addr alignment required for kernel */
 	uint8_t relocatable_kernel; /* Whether kernel is relocatable or not */
 	uint8_t min_alignment; /* Minimum alignment, as a power of two */
--- a/src/firmware/kexec.c
+++ b/src/firmware/kexec.c
@@ -35,6 +35,10 @@
 #define ALIGNUP(x, a) (((x - 1) & ~(a - 1)) + a)
 #endif
 
+#ifndef ALIGNDOWN
+#define ALIGNDOWN(x, a) (-(a) & (x))
+#endif
+
 #define BASE_GDT 0x2000ull
 #define BASE_ZEROPAGE 0x3000ull
 #define BASE_CMDLINE 0x4000ull
@@ -170,6 +174,7 @@ kexec_load_kernel(char *path, char *cmdline) {
 static int
 kexec_load_ramdisk(char *path) {
 	uint64_t ramdisk_start;
+	uint32_t initrd_max;
 	volatile struct zero_page *zp;
 	size_t sz;
 	FILE *f;
@@ -184,7 +189,18 @@ kexec_load_ramdisk(char *path) {
 	sz = (size_t) ftell(f);
 	fseek(f, 0, SEEK_SET);
 
-	ramdisk_start = ALIGNUP((kernel.base + kernel.size), 0x1000ull);
+	/* highest address for loading the initrd */
+	if (zp->setup_header.version >= 0x203) {
+		initrd_max = zp->setup_header.initrd_addr_max;
+	} else {
+		initrd_max = 0x37ffffff; /* Hardcoded value for older kernels */
+	}
+
+	if (initrd_max >= memory.size) {
+		initrd_max = ((uint32_t) memory.size - 1);
+	}
+
+	ramdisk_start = ALIGNDOWN(initrd_max - sz, 0x1000ull);
 
 	if ((ramdisk_start + sz) > memory.size) {
 		/* not enough memory */


zchee:
I don't know this change is correct, but it has the same value as the qemu without `pcmc->acpi_data_size`.

ijc25:
CI failures seem to be down to some ocaml ecosystem changes -- @djs55 we happened to be talking about this yesterday, were you going to raise a PR?

@zchee The code here looks ok, but I'm a bit concerned what will happen with >3G of RAM, when there will be a hole in the address space between 3-4GB as I mentioned in the final paragraph of https://github.com/docker/hyperkit/issues/63#issuecomment-252581435. In this case `memory.size` would perhaps point into an MMIO hole. e.g. with `memory.size` == 3.5GB of memory then the address space would be (I think) laid out like:
```
0-3GB       RAM
3-4GB       MMIO space (virtio, other devices etc)
4GB-4.5GB   RAM
```
But I think your code (assuming large enough `zp->setup_header.initrd_addr_max`) will place the initrd just below 3.5GB, which is in the middle of the MMIO hole, so badness will occur.

I think it would be acceptable to clamp the initrd address to `lowmem_limit` as well as `memory.size` (i.e. whichever is smaller). In the example above then the initrd would be loaded just below 3GB, which I think is going to be fine -- it should be way way way above anything the kernel would be expected to use during boot and is much simpler that figuring out that the real end of RAM is at 4.5GB.

zchee:
@ijc25 
> CI failures seem to be down to some ocaml ecosystem change

yeah, understood. waiting for fix, or I will debug it.

and, OK. I do not understand all yet, but got it your said meaning.
I will comment later.

zchee:
@ijc25 Now, I'm trying to understand look at the source of xhyve and qemu based on your comment.
Please wait a moment...

And, will this pull request be merged after fixing CI error?

ijc25:
> And, will this pull request be merged after fixing CI error?

I think we would need to fix the MMIO hole problem I described too (or explain why it is not a problem).

zchee:
@ijc25 Ah, Sorry.
Of curse, I'm trying to fix MMIO hole issue. Now trying.

I mean, If I fixed the MMIO issue, push to GitHub, and you said LGTM.
But this pull request merge process is after the fixed of the circlecI failed issue?

ijc25:
We need to fix the MMIO issue, rebase onto the CI fix and then review the final result, but there doesn't seem to be any obvious reason why this wouldn't be able to be merged then.

ijc25:
> rebase onto the CI fix 

I hadn't noticed that the CI is already fixed by https://github.com/docker/hyperkit/pull/64. So please rebase these changes onto the current master.

zchee:
@ijc25 OK, I understand. Sorry for confusing you..

zchee:
@ijc25 Ah, got it. I will fix MMIO issue and rebase it.

zchee:
@ijc25 Sorry again if my interpretation is wrong. I hope that I had understood you said meaning.
Also, sorry this comment is so long... :(

===

I interpreted as below:

In the case of below situation, will guest memory is split to low and high. and MMIO hole exists.

- protocol version(`zp->setup_header.version`) is `0x203` or newer
- guest memory size larger than 3GB. now assuming to `4831838208`(4.5GB)
- initrd(`sz`) size assuming to `805306368` (`4831838208 - 3221225472 / 2`)
- assuming to `if initrd_max >= memory.size` is true

The low memory has limited size that is assigned `lowmem_limit = (3ull << 30)` at the [`xh_vm_create`](https://github.com/docker/hyperkit/blob/master/src/vmm/vmm_api.c#L89-L108), actually `3221225472`(3GB)
The high memory always starts the `4294967296`(4GB).
The memory layout like

```sh
(start) 0 ~ 3221225472 ~ MMIO hole ~ 4294967296 ~ 4831838208 (end)
```

When re-calculating the `initrd_max` based on the `memory.size` **only**, You are concerned that initrd are overlapped in MMIO space. like

```c
memory.size - 1  -     sz     ramdisk_start // before aligndown
(4831838208 - 1) - 805306368 = 4026531839

ALIGNDOWN(4026531839, 0x1000ull) = 4026527744 // overlapped MMIO hole
```

So, you think we should compare the `memory.size` and `xh_vm_get_lowmem_limit()`, and use the smaller one. like
```c
#include <xhyve/support/misc.h>
.
.
		initrd_max = ((uint32_t) min(memory.size, xh_vm_get_lowmem_limit()) - 1);
```
Is this interpretation is correct?

===

If correct, I also assumed almost the same behavior.
but AFAIK, It does not happen (only tested normal boot `hyperkitrun.sh` with 1GB and 8GB)

Because, `memory.size` assigned this line,
https://github.com/docker/hyperkit/blob/master/src/firmware/kexec.c#L229

```c
	memory.size = xh_vm_get_lowmem_size();
```

and that value(`3221225472`) does not change until the loading initrd.
I do not know whether it is correct behavior(assign lowmem_size), but it will never be higher than the `lowmem_limit`, also `memory.size == 3758096384(3.5GB)` is impossibility value.
actually

```c
memory.size - 1  -     sz     ramdisk_start // before aligndown
(3221225472 - 1) - 805306368 = 2415919103

ALIGNDOWN(2415919103, 0x1000ull) = 2415915008
```

We already get the low memory size on `memory.size` and its same as `xh_vm_get_lowmem_size`.
And, I thought will degrade performance a little if call the `xh_vm_get_lowmem_size` function.
Therefore I use `memory.size` only.

===

Here is added some `fprintf` debug.
https://gist.github.com/zchee/6f12687901b18e9867367f996f93010b

I'm really sorry if I'm wrong...

ijc25:
> Is this interpretation is correct?

I think so, yes.

> Because, memory.size assigned this line,
https://github.com/docker/hyperkit/blob/master/src/firmware/kexec.c#L229

Ah, that's surprising but you are correct that having `memory.size = xh_vm_get_lowmem_size();` addresses the issue I was concerned with. Sorry for not spotting this code sooner!

So: LGTM, I'll merge. Thanks for your contribution and patience.

zchee:
@ijc25 Thanks merged!
and https://github.com/docker/hyperkit/pull/68, yeah, I think good change and easier to read.

Anyway, thanks again for code & comment review :)

PR #68, opened 2016-10-17
kexec: Rename "memory" to "lowmem".

ijc25:
Various bits of code (e.g. kernel and initrd loading) rely on memory.base and
memory.size referring to lowmem only, with highmem being handled separately, in
order to avoid locating things into the MMIO region between the two.

Rename the variable to make this more obvious and to forestall any ideas about
"fixing" memory.size.

/cc @zchee I think this would have saved us a few RTTs on #66!
/cc @rneugeba I'll hold off on this until after #65 and rebase accordingly once you have merged.

### DIFF
--- a/src/lib/firmware/kexec.c
+++ b/src/lib/firmware/kexec.c
@@ -48,7 +48,7 @@
 static struct {
 	uintptr_t base;
 	size_t size;
-} memory, kernel, ramdisk;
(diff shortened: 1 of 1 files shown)


rneugeba:
LGTM, makes the code easier to read

ijc25:
Rebased onto #65

PR #70, opened 2016-10-18
README.md: Attempt to direct downstream bug reports to downstream bug…

ijc25:
… tracker

We get a small but steady stream of Docker for Mac bug reports filled against
Hyperkit, many of which are not actually Hyperkit issues. Try and direct users
to file the bugs downstream by default so that they can be more effectively
triaged.


rneugeba:
LGTM. Merging

PR #73, opened 2016-10-23
First pass at reducing the diff between bhyve and hyperkit

rneugeba:
xhyve/hyperkit was derived from bhyve and shares more common code with it than there are differences. This is a first pass at trying to reduce the difference and mostly does two things:
- whitespace cleanups. Bring hyperkit code more inline with bhyve formatting
- clean up casts introduced in xhyve. While the casts are necessary to compile with clang, xhyve used a lot of unnecessary parentheses and whitespaces. Cleaning them up increases the chance of upstreaming them to bhyve.

/cc @avsm @luigirizzo


src/lib/acpitbl.c
```diff
@@ -163,21 +182,24 @@ acpitbl_build_rdsp(void) {
...
 	};
 
-	rdsp = (void *) (((uintptr_t) tb) + 0);
+	rdsp = (char *)tb + 0;
 	/* copy RDSP template to guest memory */
 	memcpy(rdsp, rdsp_tmpl, 36);
 	/* fixup table */
-	acpitbl_write32(rdsp, 0x10, ((uint32_t) (XHYVE_ACPI_BASE + RSDT_OFFSET)));
-	acpitbl_write64(rdsp, 0x18, ((uint64_t) (XHYVE_ACPI_BASE + XSDT_OFFSET)));
+	acpitbl_write32(rdsp, 0x10, (uint32_t)XHYVE_ACPI_BASE + RSDT_OFFSET);
```
ijc25: This ones a bit subtle, depending on how the constants are defined. If `RSDT_OFFSET` were `0xXXXXUL` (which wouldn't be unreasonable) then I think the `uint32_t + unsigned long` would potentially be promoted to an `unsigned long`, despite the first operand having been cast to `uint32_t`, which wasn't quite as intended.

We could write it as `(unt32_t)(FOO + BAR)` (i.e. just fix the whitespace not the braces) but perhaps we would be better off letting the implicit conversion (due to the argument to `acpitbl_write32` taking a `uint32_t`) DTRT? That might mean changing the constants to be unsigned, but arguably that would be correct in its own right?

ijc25: BTW, I would be fine with a "lets apply as is now and reconsider later" answer here...

rneugeba: I'll fix it. was a bit too eager removing parentheses

src/lib/block_if.c
```diff
@@ -413,22 +431,29 @@ blockif_proc(struct blockif_ctxt *bc, struct blockif_elem *be, uint8_t *buf)
...
+			err = EROFS;
+		else if (bc->bc_ischr) {
+			arg[0] = br->br_offset;
+			arg[1] = br->br_resid;
+			if (ioctl(bc->bc_fd, DIOCGDELETE, arg))
+				err = errno;
+			else
+				br->br_resid = 0;
+		} else
+#else
```
ijc25: Shall we just delete this? We can always get it back from git if we need it..

rneugeba: was contemplating this as well. Yeah, I'll remove it

src/lib/pci_virtio_net_tap.c
```diff
@@ -56,34 +56,39 @@
 #define VTNET_RINGSZ 1024
 #define VTNET_MAXSEGS 32
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-macros"
```
ijc25: I think we should just add this to `CFLAGS`

ijc25:
LGTM, I made some comments which were mostly asides rather than issues with the changes. Only one worth thinking about is turning off `Wunused-macros` globally via CFLAGS rather than file by file.

> While the casts are necessary to compile with clang

I didn't spot any casts which didn't look like good practice (as opposed to just placating the compiler), but isn't FreeBSD (and hence bhyve) compiling with clang by default these days too? Is it just that Apple's version of clang is defaulting to more warnings, or is it that Hyperkit's `config.mk` is turning on lots of extra stuff? Just curious why these casts haven't already been found necessary in bhyve.

src/lib/pci_emul.c
```diff
@@ -572,7 +572,7 @@ pci_emul_alloc_pbar(struct pci_devinst *pdi, int idx, uint64_t hostbase,
 	limit = 0;
 
 	if ((size & (size - 1)) != 0)
-		size = 1UL << flsl((long) size); /* round up to a power of 2 */
+		size = 1UL << flsl((long)size); /* round up to a power of 2 */
```
rneugeba: I think the signed-ness of the variable should not matter for `flsl` as it would just look at the bits. `long` is indeed 64bit so this should work...though i'm going to change this to `long long` which is a bit more correct, I think, at least on 32bit systems

src/lib/block_if.c
```diff
@@ -82,33 +83,33 @@ enum blockstat {
...
-	off_t be_block;
+	struct blockif_req  *be_req;
+	enum blockop	     be_op;
+	enum blockstat	     be_status;
+	pthread_t            be_tid;
+	off_t		     be_block;
 };
 
 struct blockif_ctxt {
-	int bc_magic;
```
avsm: do we know why an assert is appropriate here instead of a dynamic check? All the lseeks appear to assume success.

src/lib/block_if.c
```diff
@@ -372,7 +389,7 @@ blockif_proc(struct blockif_ctxt *bc, struct blockif_elem *be, uint8_t *buf)
```
avsm: would just prefer to keep this codebase clean without the "bhyve removal" comments -- its pretty obvious from the perror that alternative implementations could exist.

grehan-freebsd:
>Just curious why these casts haven't already been found necessary in bhyve

usr.sbin/bhyve is built with WARNS=2 in the Makefile - this may not be as strict as the default with clang on OSX.

More than happy to take changes which would allow a stricter WARNS value to be used on FreeBSD.

src/lib/block_if.c
rneugeba: I added the comment, because in a previous version i had `#if 0` and the original version had the bhyve code commented out. Both of which make it blindingly obvious that something from the original source was hacked to make it work on OSX. I'm for removing dead code, but do prefer to leave a breadcrumb indicating that something may have changed due to a hack, just in case someone is debugging this code in a few months time.

rneugeba:
@grehan-freebsd hyperkit is compiled with `-Weverything` and then some exceptions added afterwards. Some/most of the casts seem like good practise (like not assuming a `void *` points to something of size 1 byte, unsigned vs signed arithmetic, assignments to smaller data types etc)

I currently don't have a FreeBSD dev environment set up, but we are definitely interested in reducing the diff, especially for the device emulation portions. It might take a few more passes before we get there, re @luigirizzo email to  emulation@freebsd.org

src/lib/block_if.c
rneugeba: I think I like that, ie either `__FreeBSD__` or feature (if appropriate). Will update again...though it looks like GH was royally confused with my force push which also updated the author and committer. Might have to open a new PR

luigirizzo:
Peter, Rolf: I am in favour of strict compiler warnings, the only one I would remove in CFLAGS is -Wpadded because it seems to make life miserable: it complains both when a struct is declared __packed but no padding would be inserted anyways, and when it is not __packed but padding is inserted. The whole point of using __packed is tell the compiler which structs should be packed and which should not, then leave us alone.

luigirizzo:
Regarding diff reduction with bhyve, given that the task is much larger than I expected, I would approach it in an opportunistic way, i.e. as we go through a subsystem (e.g. for merging a new device emulation or fixing a feature) also take the chance to revert the gratuitous changes applied in the port (whitespace, file renaming, changes in function prototypes). Likewise for importing changes into FreeBSD.

ijc25:
@luigirizzo I couldn't agree more regarding `-Wpadded`!

src/lib/block_if.c
avsm: It looks like everything got very confused after a force push...

thebsdbox:
Given most of the bhyve code is focussed on contexts of a number of VMs, would you look at reinstating the code to act in the following manner (although always referencing the same context as hyperkit doesn't run multiple VMs the same way bhyve does). This would make it much simpler to import the additional pci devices that have been added into bhyve e.g.
- E1000
- VNC driver
   - USB device emul (needed by vnc for mouse/kb) <- this being the worst to rely on numerous vm contexts.

(just a thought)

rneugeba:
I haven't looked at the context structure in FreeBSD in too much detail yet, but re-instating it would maybe make sense. However, there are q fair few number of lower hanging fruits, which I would pick first

rneugeba:
closing this pull request. GH got confused with rewriting the git authot. Will open an updated one shortly.

PR #74, opened 2016-10-25
Link in libev, which is sometimes needed for Lwt.

yallop:
[Lwt](https://github.com/ocsigen/lwt) [optionally depends on libev](https://github.com/ocsigen/lwt/blob/d638502667005270cb69bad282f89afa9370838e/opam#L43).  This PR updates the Makefile to statically link in `libev` if it's available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,9 @@ HAVE_OCAML_QCOW := $(shell if ocamlfind query qcow uri >/dev/null 2>/dev/null ; 
 ifeq ($(HAVE_OCAML_QCOW),YES)
 CFLAGS += -DHAVE_OCAML=1 -DHAVE_OCAML_QCOW=1 -DHAVE_OCAML=1
 
+LIBEV_FILE=/usr/local/lib/libev.a
+LIBEV=$(shell if test -e $(LIBEV_FILE) ; then echo $(LIBEV_FILE) ; fi )
+
 # prefix vsock file names if PRI_ADDR_PREFIX
 # is defined. (not applied to aliases)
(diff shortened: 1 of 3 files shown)


avsm:
Could we also add `libev` to the `brew install opam` section in the `README`, so that we encourage use of this engine by default?

yallop:
@avsm: good idea; see fad3e85.

ijc25:
Can we also add it to `circle.yml` so it gets installed for CI too please.

Is libev not pulled in by `opam depext`? (or could/should it be).

Don't forget the Signed-of-by and since you need to force push to do that you might as well squash it all into one commit...

avsm:
> Is libev not pulled in by opam depext

It will be pulled in by `opam depext` if `opam depext -ui conf-libev lwt` is used to invoke the installation.  `conf-libev` is an optional dependency in the OPAM Lwt metadata, and so it must be explicitly specified in the installation to activate it.

yallop:
I've updated `circle.yml`, signed, squashed and pushed.

ijc25:
> It will be pulled in by opam depext if opam depext -ui conf-libev lwt is used to invoke the installation. conf-libev is an optional dependency in the OPAM Lwt metadata, and so it must be explicitly specified in the installation to activate it.

Not sure if that is that better or worse than invoking `brew` direct? Maybe slightly preferable to go all-opam-all-the-time for ocaml realted deps?

avsm:
This lgtm -- we can replace the `opam install` instructions with `opam depext` but probably not in this PR, as it affects the CircleCI caching if we squash multiple steps into a single one.

avsm:
thanks!

PR #75, opened 2016-11-02
First pass at reducing the diff between bhyve and hyperkit

rneugeba:
xhyve/hyperkit was derived from bhyve and shares more common code with it than there are differences. This is a first pass at trying to reduce the difference and mostly does two things:

- clean up casts introduced in xhyve. While the casts are necessary to compile with clang, xhyve used a lot of unnecessary parentheses and whitespaces. Cleaning them up increases the chance of upstreaming them to bhyve.
- add `-Wno-padded` to `CFLAGS` and remove pragmas from the source code
- mark commented out FreeBSD code with `#ifdef __FreeBSD__`
- whitespace cleanups. Bring hyperkit code more inline with bhyve formatting

/cc @ijc25 @avsm @luigirizzo

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -40,6 +40,7 @@ CFLAGS_WARN := \
   \
   -Wno-dollar-in-identifier-extension \
   -Wno-gnu-statement-expression \
+  -Wno-packed \
   -Wno-padded \
   -Wno-reserved-id-macro \
   -Wno-unknown-warning-option \
--- a/config.mk
+++ b/config.mk
@@ -40,6 +40,7 @@ CFLAGS_WARN := \
   \
   -Wno-dollar-in-identifier-extension \
   -Wno-gnu-statement-expression \
+  -Wno-padded \
   -Wno-reserved-id-macro \
   -Wno-unknown-warning-option \
   -Wno-unused-macros
--- a/config.mk
+++ b/config.mk
@@ -32,14 +32,17 @@ CFLAGS_OPT := \
   -flto \
   -fstrict-aliasing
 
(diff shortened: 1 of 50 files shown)


config.mk
```diff
@@ -32,14 +32,18 @@ CFLAGS_OPT := \
...
-  -Wno-reserved-id-macro \
+  -pedantic \
+  \
   -Wno-dollar-in-identifier-extension \
   -Wno-gnu-statement-expression \
-  -pedantic
+  -Wno-padded \
+  -Wno-reserved-id-macro \
+  -Wno-unknown-warning-option \
+  -Wno-unused-macros
```
ijc25: Would be nice to eventually have a `cc-option` make macro, but `-Wno-unknown-warning-option` will suffice for now.

ijc25:
Given #73 shouldn't this be titled "second pass..." :trollface: 

Anyway, looks like the comments from #73 have all been address and therefore LGTM.

config.mk
rneugeba: I was too considering to remove the packed pragmas and add `-Wno-packed`. seems superfluous to me.

@ijc25 any strong view on this? If not I'll update the PR with another patch

ijc25: A weak preference for both `-Wno-packed` and `-Wno-padded`, IOW go ahead...

rneugeba:
force pushed to also add `-Wno-packed` and removed the pragmas around packed strutures

ijc25:
`-Wno-packed` addition LGTM also, merging.

PR #76, opened 2016-11-03
Partial support for raw devices

johanneswuerbach:
I haven't checked, whether all kinds of volumes work, but virtual disk images do, which allows you to use sparse volumes now with xhyve.

Example:

```
hdiutil create -megabytes 20 -fs "MS-DOS" disk
hdiutil attach disk.dmg // Get disk number
xhyve ... -s 4:0,ahci-hd,/dev/rdiskN ...
```

Ported from https://github.com/mist64/xhyve/pull/80

### DIFF
--- a/src/lib/block_if.c
+++ b/src/lib/block_if.c
@@ -535,7 +535,7 @@ blockif_open(const char *optstr, const char *ident)
 	struct blockif_ctxt *bc;
 	struct stat sbuf;
 	// struct diocgattr_arg arg;
-	off_t size, psectsz, psectoff;
+	off_t size, psectsz, psectoff, blocks;
 	int extra, fd, i, sectsz;
 	int nocache, sync, ro, candelete, geom, ssopt, pssopt;
 	mirage_block_handle mbh;
@@ -653,8 +653,17 @@ blockif_open(const char *optstr, const char *ident)
 		if (ioctl(fd, DIOCGPROVIDERNAME, name) == 0)
 			geom = 1;
 #else
-		perror("xhyve: raw device support unimplemented");
-		goto err;
+		blocks = 0;
+		if (ioctl(fd, DKIOCGETBLOCKCOUNT, &blocks) < 0 ||
+			ioctl(fd, DKIOCGETBLOCKSIZE, &sectsz)) {
+			perror("Could not fetch dev blk/sector size");
+			goto err;
+		}
+
+		assert(blocks != 0);
+		assert(sectsz != 0);
+
+		size = blocks * sectsz;
 #endif
 	} else
 		psectsz = sbuf.st_blksize;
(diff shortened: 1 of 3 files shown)


src/lib/block_if.c
```diff
@@ -618,17 +618,17 @@ blockif_open(const char *optstr, const char *ident)
...
-		// }
-		// assert(size != 0);
-		// assert(sectsz != 0);
+		if (ioctl(fd, DKIOCGETBLOCKCOUNT, &blocks) < 0 ||
+        ioctl(fd, DKIOCGETBLOCKSIZE, &sectsz)) {
+		  perror("Could not fetch dev blk/sector size");
+		 	goto err;
+		}
+		assert(blocks != 0);
+		assert(sectsz != 0);
```
avsm: These should probably be dynamic checks that also `perror`, in case we compile with asserts disabled.

ijc25: Since they were commented out before I presume these originally come from bhyve and it would be better not to diverge on such trivial things. Maybe bhyve could be changed too though.

To that end, rebasing this PR onto #75 seems like a good idea to me.

src/lib/block_if.c
```diff
@@ -618,17 +618,17 @@ blockif_open(const char *optstr, const char *ident)
 	sectsz = DEV_BSIZE;
 	psectsz = psectoff = 0;
 	candelete = geom = 0;
+  blocks = 0;
```
ijc25: Please indent to match the surrounding code (and again in the following if/ioctl combination).

ijc25:
Nice work. Apart from my comments on the code it would be good if the commit message incorporated the info from the PR text too since it is useful reference information IMHO.

I expect `hyperkit.1` probably needs updating to reflect this change too.

src/lib/block_if.c
avsm: Agreed its not worth the divergence.

src/lib/block_if.c
johanneswuerbach: Sorry, editor fail :-(

johanneswuerbach:
Regarding the `hyperkit.1 `, I just modified the last example (could also create a new one) The other parts seem to already mention using devices like `/dev/disk` even when is wasn't working.

ijc25:
LGTM but please rebase onto master now that #75 has been merged (should make your patch a lot smaller too, which is good!)

ijc25:
LGTM but please rebase onto master now that #75 has been merged (should make your patch a lot smaller too, which is good!)

johanneswuerbach:
Rebased 👍

johanneswuerbach:
Added a small test in a separate commit, wdyt?

ijc25:
LGTM, merging, thanks

PR #77, opened 2016-11-07
Dump VM state on VMEXIT_ABORT

ijc25:
Otherwise if we take a trap for e.g. VM_EXITCODE_VMX then we have very little
to go on (basically only RIP and exit_reason + qualification).

Plumb the existing hvdump function (which dumps a lot of useful VMCS state,
including guest register state) up through the various API layers and have the
top-level call it for any VMEXIT_ABORT.

/cc @rneugeba you might have opinions on the API layering and how to expose things?

Tested by adding an artificial VMEXIT_ABORT to `vcpu_loop` after 100 exits since I don't currently have an easy way to force such an exit.

### DIFF
--- a/src/hyperkit.c
+++ b/src/hyperkit.c
@@ -633,6 +633,7 @@ vcpu_loop(int vcpu, uint64_t startrip)
 		case VMEXIT_CONTINUE:
 			break;
 		case VMEXIT_ABORT:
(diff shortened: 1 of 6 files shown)


rneugeba:
LGTM, merging
The layering of calls is indeed a bit tedious but best to follow the pattern that's currently there and clean up the layering later for all the calls.

PR #78, opened 2016-11-24
vsock: do not clear s->credit_update_required if no traffic sent

ijc25:
pci_vtsock_rx_thread() assumes that if pci_vtsock_proc_rx() returned >= 0 then
something will have been queued on the ring and therefore a credit update is no
longer required (credit updates can piggy back on any traffic).

However there were two paths through pci_vtsock_proc_rx() which did not send
anything:

- The peer has no buffer space available (when peer_buf_alloc - (rx_cnt -
  peer_fwd_cnt) == 0).
- Read on the Unix domain socket returned EAGAIN (nothing to read).

This would result in rx_thread() clearing s->credit_update_required when the
update was still required, which could lead to a deadlock for that socket if
there happened to be no other pending tx traffic to generate a second attempt
at a credit update.

Fix this by:

- moving the clearing of the flag to exactly the place which enqueues the
  response (corresponding to the calls to vq_relchain() in proc_rx())
- ensuring that the two cases above send an explicit OP_CREDIT_UPDATE if a
  credit update is needed (this is less complex than a separate return code and
  having the caller fall-thru to its credit_update_required case.

With these changes rx_thread() can now assert that no credit update is required
after proc_rx() returns >= 0.

### DIFF
--- a/src/lib/pci_virtio_sock.c
+++ b/src/lib/pci_virtio_sock.c
@@ -1612,9 +1612,13 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
 
 /*
  * Returns:
- *  -1 == no descriptors available
+ *  -1 == no descriptors available, nothing sent, s->credit_update_required untouched
  *   0 == nothing done (sock has shutdown, peer has no buffers, nothing on Unix socket)
  *  >0 == number of bytes read
+ *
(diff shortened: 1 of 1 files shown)


rneugeba:
LGTM. Thanks for the detailed description of the issue and fix

PR #79, opened 2016-11-25
fixed README's code blocks

lihuanshuai:
these code blocks in README did not work the right way, I just fixed them

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,12 @@ Notes:
   hyperkit so the build will find the ocaml environment.
 - Any previous pin of `mirage-block-unix` or `qcow-format`
   should be removed with the commands:
-
-    $ opam update
-    $ opam pin remove mirage-block-unix
-    $ opam pin remove qcow-format
+  
+  ```sh
+  $ opam update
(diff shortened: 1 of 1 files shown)


ijc25:
Thanks. I wonder why the three code blocks further up the file which look to be identically indented to this one render fine in https://github.com/docker/hyperkit/blob/master/README.md.

In any case perhaps we should markup all of the code blocks using the "```sh" syntax?

rneugeba:
Thanks for the fix.

ijc25:
BTW, the reason for the difference is that it immediately follows a bulleted list. Seems like it could also have been fixed by indenting by 8 spaces instead of 4 (4 for the bullet list and another 4 for the fixed formatting).

lihuanshuai:
@ijc25 you are right. in fact, 6 spaces is enough for  this case, but an indention with more 4 space is not a good choice in markdown.

rneugeba:
apologies, was a bit to trigger happy merging this. The resulting README.md looked fine, but it would be good to fix this properly

PR #80, opened 2016-11-29
Vsock locking fixes

ijc25:
Various issues found by running hvgostress from https://github.com/rneugeba/virtsock and from investigations of https://github.com/docker/for-mac/issues/825 (although sadly even these do not fix that issue).


rneugeba:
I have to admit that I don't understand the intricacies of the code, but:
- your detailed descriptions seem to match the patches
- I trust you ran the stress test program and it passes with the fixes

Based on this, LGTM

PR #81, opened 2016-12-02
Add support for qcow delete/discard/TRIM

djs55:
This PR adds support for `delete` (aka `discard` or `TRIM`) via the Mirage Qcow library.

The capability `candelete` (used to set `ATA_SUPPORT_RZAT | ATA_SUPPORT_DRAT` in `src/lib/pci_ahci.c`) is returned from an expanded `mirage_block_stat` call, while the `delete` operation itself is plumbed through a new function `mirage_block_delete`.

Since the Mirage Qcow library supports multiple configuration options (and more are going to be added over time), an extra configuration string is plumbed through via the new block parameter `qcow-config=`.

Related to [docker/for-mac#371]

### DIFF
--- a/src/lib/block_if.c
+++ b/src/lib/block_if.c
@@ -555,6 +555,7 @@ blockif_open(const char *optstr, const char *ident)
 	char *nopt, *xopts, *cp;
 	struct blockif_ctxt *bc;
 	struct stat sbuf;
+	struct mirage_block_stat msbuf;
 	// struct diocgattr_arg arg;
 	off_t size, psectsz, psectoff, blocks;
 	int extra, fd, i, sectsz;
@@ -611,6 +612,8 @@ blockif_open(const char *optstr, const char *ident)
 	if (sync)
 		extra |= O_SYNC;
 
+	candelete = 0;
+
 	if (use_mirage) {
 #ifdef HAVE_OCAML_QCOW
 		mirage_block_register_thread();
@@ -619,11 +622,11 @@ blockif_open(const char *optstr, const char *ident)
 			perror("Could not open mirage-block device");
 			goto err;
 		}
-
-		if (mirage_block_stat(mbh, &sbuf) < 0) {
+		if (mirage_block_stat(mbh, &sbuf, &msbuf) < 0) {
 			perror("Could not stat backing file");
 			goto err;
 		}
+		candelete = msbuf.candelete;
 #else
 		abort();
 #endif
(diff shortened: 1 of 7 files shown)


ijc25:
> As the Mirage change is not yet released upstream (although I'm working on it) I've temporarily added back an opam pin to a specific branch.

I can't see that here, just the name change from `qcow-format` to `qcow`.

When updating `circle.yml` please also update the for-human steps in `README.md`.

djs55:
Sorry I rebased this morning after @yallop merged the `qcow.0.6.0` upstream so I removed the pin but forgot to update the text.

src/lib/block_if.c
```diff
@@ -427,23 +459,11 @@ blockif_proc(struct blockif_ctxt *bc, struct blockif_elem *be, uint8_t *buf)
...
-			else
-				br->br_resid = 0;
-		} else
-			err = EOPNOTSUPP;
-#else
-		err = EOPNOTSUPP;
-#endif
+		if (block_delete(bc, br->br_offset, br->br_resid) < 0) {
+			err = errno;
+			break;
```
rneugeba: I'm in two minds about this. In general, I don't like the these very long case statement and prefer the approach you have taken (calling a function). It does however make it harder to diff against upstream bhyve...

@ijc25 any preference?

src/lib/mirage_block_c.c
```diff
@@ -86,29 +87,31 @@ ocaml_mirage_block_stat(mirage_block_handle h, struct stat *out, int *err) {
...
-		out->st_nlink = 1;
-		out->st_uid = 0;
-		out->st_gid = 0;
-		out->st_rdev = 0;
-		out->st_size = (off_t)(sector_size * size_sectors);
-		out->st_blocks = (blkcnt_t)size_sectors;
-		out->st_blksize = (blksize_t)sector_size;
-		out->st_flags = 0;
-		out->st_gen = 0;
+		bzero(out->stat, sizeof(struct mirage_block_stat));
```
ijc25: Or another alternative would be to not embed the `stat` into the `mirage_block_stat` at all and simply add a new argument to this function for the mirage specific stuff.

ijc25:
Please could you rebase onto master which now contains #82 so we can be sure this still at least builds both with and without ocaml.

djs55:
OK I've
- rebased: the CI looks happy
- changed `mirage_block_stat` to have an extra parameter parameter
- fixed the indenting around the stray `else`
- replaced `qcow-format` with `qcow` in the README.md and ran through it as a sanity check
- smoke tested it again just in case :)

I think the only outstanding question is @rneugeba 's about moving the `BOP_DELETE` handling into a `block_delete` function rather than leaving it inline.

src/lib/block_if.c
ijc25: Looks like we already took the approach with flush of pulling it out so we may as well do so here too, but I agree wrt diffing against bhyve.

I think the `// off_t arg[2];` at the top of `blockif_proc` can/should now be removed though since this moves the vestigial reference to it elsewhere.

djs55: OK I've kept the function but have removed the `// off_t arg[2]`. CI is happy again (not that I expected it to object!)

ijc25:
LGTM.

PR #82, opened 2016-12-05
circle.yml: Build and test with and without ocaml

ijc25:
First build and test (which assumes that ocaml is not present in the base
circle image) then install opam etc and do it again with ocaml enabled.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,6 +7,9 @@ machine:
     version: "7.3"
 test:
   override:
+    - make clean
+    - make all
+    - make test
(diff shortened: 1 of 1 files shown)


ijc25:
circle failed the `opam init --yes` test step with `curl: (6) Could not resolve host: opam.ocaml.org`, but then went on to build and test without ocaml. It did at least fail things overall though. Since this is unrelated I'll hit retry.

rneugeba:
LGTM

PR #84, opened 2016-12-06
vsock: correctly account partial writes on tx

ijc25:
When buffering due to a partially successful write we need to add the
successful part of the write to fwd_cnt otherwise we miss those bytes in our
accounting and will eventually fall behind enough that we think we cannot
forward any more traffic.

With this in place the mysql test case in https://github.com/docker/for-mac/issues/825 now succeeds for me.


rneugeba:
LGTM

ijc25:
Thanks for the very fast review! Merging.

PR #86, opened 2016-12-07
ocaml: initialise the logging system used by the qcow library

djs55:
Previously the qcow library could emit errors and warnings but these would be dropped since the logging system wasn't initialised. This patch should cause these messages to be printed to stderr, along with the other output from hyperkit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ FIRMWARE_LIB_SRC := \
 
 HYPERKIT_SRC := src/hyperkit.c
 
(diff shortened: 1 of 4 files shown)


ijc25:
Will you need to change `circle.yml` too?

djs55:
Oh yes good point, pushed an updated patch.

ijc25:
Looks like the new modules were already pulled in by some opam dependency, but perhaps this should either be explicit in both `README.md` and `circle.yml` or implicit in both. FWIW I favour the former...

ijc25:
Heh, looks like we raced! Thanks.

ijc25:
LGTM subject to CI thumbs up.

djs55:
Thanks!

PR #87, opened 2016-12-07
Add a missing '='.

yallop:
Signed-off-by: Jeremy Yallop <yallop@docker.com>


ijc25:
Ahem, well that was one way of ensuring the assertion was true!

Thanks, LGTM.

PR #89, opened 2016-12-09
ocaml: qcow-config should be optional

### LINKED ISSUE #88: Hyperkit segfaults if format=qcow is passed to it without qcow-config
Hyperkit segfaults if a qcow disk is used, but the `qcow-config` option is not explicitly passed to it.

Here's an example:

```
$ FLAGS="-A -s 0:0,hostbridge -s 31,lpc -l com1,stdio -m 4G -s 2:0,virtio-net"
```

This segfaults:
```
# com.docker.hyperkit $FLAGS -s "4,virtio-blk,file:///tmp/test.qcow2,format=qcow" -f "kexec,vmlinuz,initramfs.img,earlyprintk=serial console=ttyS0"
```

While this doesn't:
```
# com.docker.hyperkit $FLAGS -s "4,virtio-blk,file:///tmp/test.qcow2,format=qcow,qcow-config=" -f "kexec,vmlinuz,initramfs.img,earlyprintk=serial console=ttyS0"
```

I'm not sure if `qcow-config` is intended to be a mandatory option, but if it is, hyperkit should probably throw an error instead of segfaulting.

djs55:
This patch fixes a segfault triggered by setting `format=qcow` but not providing a `qcow-config` parameter. The Mirage qcow interface already treats this configuration as an optional value. This patch modifies the C -> OCaml interface to map NULL onto None.

Fixes #88

### DIFF
--- a/src/lib/mirage_block_c.c
+++ b/src/lib/mirage_block_c.c
@@ -48,11 +48,17 @@ if (fn == NULL) { \
 static void
 ocaml_mirage_block_open(const char *config, const char *options, int *out, int *err) {
 	CAMLparam0();
-	CAMLlocal3(ocaml_config, ocaml_options, handle);
+	CAMLlocal4(ocaml_config, ocaml_options_opt, ocaml_string, handle);
 	ocaml_config = caml_copy_string(config);
-	ocaml_options = caml_copy_string(options);
+	if (options == NULL) {
+		ocaml_options_opt = Val_int(0); /* None */
+	} else {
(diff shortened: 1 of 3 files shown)


ijc25:
LGTM (but then so did the original patch :-P)

djs55:
Thanks :) I wonder if I could add some qcow2-based test cases?

ijc25:
That would be awesome of course!

djs55:
I'll merge this and make a separate PR with test cases.

PR #90, opened 2016-12-09
Add some qcow-related tests

djs55:
- verify that we can boot up a VM with a .qcow2-backed block device and successfully write to it
- check that the without-qcow build really is without-qcow: this will fail if the CI base image were ever to have the qcow dependencies installed in it

Tests #88

### DIFF
--- a/test/test_linux_qcow.exp
+++ b/test/test_linux_qcow.exp
@@ -1,0 +1,37 @@
+#!/usr/bin/env expect
+
+set KERNEL "./vmlinuz"
+set INITRD "./initrd.gz"
+set CMDLINE "earlyprintk=serial console=ttyS0"
+
+exec qcow-tool create --size 1GiB disk.qcow2
+set PWD [ exec pwd ]
+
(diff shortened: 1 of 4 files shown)


test/test_linux_qcow.exp
```diff
@@ -0,0 +1,37 @@
...
+
+spawn ../build/com.docker.hyperkit -A -m 512M -s 0:0,hostbridge -s 31,lpc -l com1,stdio -s 1:0,ahci-hd,file:///$PWD/disk.qcow2,format=qcow -f kexec,$KERNEL,$INITRD,$CMDLINE
+set pid [exp_pid]
+set timeout 20
+
+expect {
+  timeout {puts "FAIL boot"; exec kill -9 $pid; exit 1}
+  "\r\ntc@box:~$ "
+}
+send "sudo echo 'This is a block device' > /dev/sda\r\n"
```
ijc25: Either the `sudo` isn't needed to open `/dev/sda` or this is broken because only the `echo` will run as root, the redirect `> /dev/sda` will be run by the shell as the current user.

I think the console will be a root shell, so you probably don't need any of the `sudo` do you?

If not then a common workaround is `echo 'Foo' | sudo dd of=/dev/sda` or `echo 'Foo' | sudo tee -a /dev/sda` or similar.

djs55: Good point -- it does indeed work fine without the `sudo`.

ijc25:
LGTM

djs55:
Thanks!

PR #91, opened 2016-12-09
ocaml: if a Mirage thread fails, ensure we fail the request

djs55:
The Mirage I/O operations are supposed to return a result type as follows:

    [ `Ok of result | `Error of error description ]

If the implementation actually fails i.e. by raising an exception, this would
propagate and prevent the client request ever returning. This patch catches
stray exceptions, logs something about them, and returns an error to the
caller.

### DIFF
--- a/src/lib/mirage_block_ocaml.ml
+++ b/src/lib/mirage_block_ocaml.ml
@@ -283,79 +283,90 @@ let process_one t =
   let open Protocol in
   let open Mirage_block.Error.Monad in
   let open Mirage_block.Error.Monad.Infix in
-  let result_t = match t.request with
-    | Request.Connect (block_config, qcow_config) ->
-      Block.of_config block_config
-      >>= fun base ->
-      Qcow.connect ?config:qcow_config base
-      >>= fun block ->
-      ( let open Lwt in
-        Qcow.get_info block
-        >>= fun info ->
-        return (`Ok info) )
-      >>= fun info ->
-      let h = Handle.register { Handle.block; base; info } in
-      return (Response.Connect h)
-    | Request.Get_info h ->
-      let t = Handle.find_or_quit h in
-      let open Lwt in
-      Qcow.get_info t.Handle.block
-      >>= fun info ->
-      let config = Qcow.to_config t.Handle.block in
-      let candelete = config.Qcow.Config.discard in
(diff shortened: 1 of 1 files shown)


ijc25:
```
$ git fetch origin pull/91/head
From github.com:docker/hyperkit
 * branch            refs/pull/91/head -> FETCH_HEAD
$ git diff -b origin/master FETCH_HEAD -- src/lib/mirage_block_ocaml.ml 
diff --git a/src/lib/mirage_block_ocaml.ml b/src/lib/mirage_block_ocaml.ml
index f9554eb..49b43a0 100644
--- a/src/lib/mirage_block_ocaml.ml
+++ b/src/lib/mirage_block_ocaml.ml
@@ -283,7 +283,10 @@ let process_one t =
   let open Protocol in
   let open Mirage_block.Error.Monad in
   let open Mirage_block.Error.Monad.Infix in
-  let result_t = match t.request with
+  let result_t =
+    Lwt.catch
+      (fun () ->
+        match t.request with
           | Request.Connect (block_config, qcow_config) ->
             Block.of_config block_config
             >>= fun base ->
@@ -350,12 +353,20 @@ let process_one t =
             let n = Int64.div len sector_size in
             Qcow.discard t.Handle.block ~sector ~n ()
             >>= fun () ->
-      return (Response.Delete)
+            return Response.Delete
           | Request.Flush h ->
             let t = Handle.find_or_quit h in
             Block.flush t.Handle.base
             >>= fun () ->
-      return Response.Flush in
+            return Response.Flush
+      )
+      (fun e ->
+        (* If the thread fails unexpectedly then convert this into an Error return
+           to guarantee we return an error code to the block interface instead of
+           dropping the request on the floor. *)
+        Log.err (fun f -> f "Mirage block device raised exception: %s" (Printexc.to_string e));
+        Lwt.return (`Error (`Unknown (Printexc.to_string e)))
+      ) in
   let open Lwt in
   result_t >>= fun result ->
   Ivar.fill t.ivar result;
```

LGTM, I also had a quick look through the resulting code which looks fine also.

djs55:
Thanks! I just realised this branch doesn't contain the new CI tests. I'll rebase onto latest master just in case.

ijc25:
Good idea.

djs55:
OK, CI with the new tests is happy.

PR #92, opened 2016-12-10
vsock: defer credit updates

dsheets:
This patchset introduces two new and related features:

1. Per-vsock run-time configuration by host processes via a UNIX domain socket attached fd channel
2. Limit-based credit update elision which significantly reduces the number of messages and interrupts for RPC-like protocols

The credit limit feature relies on the run-time socket configuration feature for opt-in.

In casual benchmarking, this reduces HyperKit CPU usage by 10-15% and increases throughput by ~10% under a small sequential write workload.

The configuration system is described in the commit message of af56e12:

> If messages with attached fds are sent to an AF_UNIX socket that proxies
an AF_VSOCK connection, they are treated as configuration RPC
requests. RX thread activity will block until the configuration request
is applied. The passed fd is valid for a single use in which the
requesting process writes a configuration message and, if appropriate, a
reply is written back. Once the configuration request is processed, the
fd is closed.
>
> This commit introduces version 1 of the configuration protocol that
currently has a message format that looks like:
```c
struct vtsock_config_hdr {
       uint32_t len;
       uint32_t version;
       uint32_t type;
};
```

The credit limit system is described in the commit message of 7d576c7:

> This introduces the sock->credit_cnt field that tracks the device's
understanding of the driver's peer_fwd_cnt (sock->fwd_cnt). When a
credit update may be necessary, the difference between the device's
sock->fwd_cnt and sock->credit_cnt is compared to sock->credit_limit
and, if this safe disparity is sufficiently small, no credit update is
performed and the driver believes that it has less remote buffer than it
actually does. A credit limit of zero maintains the previous behavior of
sending credit updates frequently.
>
> As normal data packets (RW) contain buf_alloc and fwd_cnt fields, this
credit limit functionality can completely elide credit updates for
sequential RPC protocols.
>
> In the case of an explicitly requested credit update, the
sock->credit_cnt is fudged to immediately trigger repayment via a credit
update message.

Please review carefully.

### DIFF
--- a/src/lib/pci_virtio_sock.c
+++ b/src/lib/pci_virtio_sock.c
@@ -459,6 +459,11 @@ static size_t iovec_clip(struct iovec **iov, int *iov_len, size_t bytes)
 	return ret;
 }
 
+/*
+ * Removes the first byte in the provided iovec and shifts the data in
+ * the rest of the iovec down (left, toward zero) by a single
+ * byte. The iovec metadata and its length remained unchanged.
+ */
 static size_t iovec_shift_left1(struct iovec **iov, int *iov_len)
 {
 	char *base;
@@ -1883,7 +1888,9 @@ static void handle_socket_config(struct pci_vtsock_sock *s, int fd)
  * Check for an attached socket configuration fd, s->mtx must be held
  *
  * If a configuration fd is attached, this will block until the
- * configuration request is handled!
+ * configuration request is handled! A configuration fd may only be
+ * attached at the beginning of the data stream and must be
+ * accompanied by a single byte written.
  *
  * Returns > 0 for regular messages, 0 for EOF, -1 for error, and -2
  * for configuration messages.
@@ -1915,6 +1922,14 @@ static ssize_t readrx(struct pci_vtsock_sock *s,
 			handle_socket_config(s, fd);
 			close(fd);
 			if (len > 1) {
+				/* If recvmsg has coalesced messages,
+				 * we must discard the configuration
+				 * fd carrier byte. Because the iovec
+				 * points to guest-shared vring
+				 * memory, we must shift the data in
+				 * the iovec rather than adjusting the
+				 * iovec metadata.
+				 */
 				iovec_shift_left1(&iov, &iov_len);
 				len--;
 			} else {
--- a/src/lib/pci_virtio_sock.c
+++ b/src/lib/pci_virtio_sock.c
@@ -1702,6 +1702,32 @@ static void apply_socket_config_set_buf_alloc(struct pci_vtsock_sock *s,
 	fprintf(stderr, "config: %s buf_alloc = %d\n", s->name, buf_alloc);
 }
 
+static int is_socket_config_malformed(struct vtsock_config_hdr *hdr)
+{
+	int rc = 0;
+	size_t hdr_sz = sizeof(struct vtsock_config_hdr);
+
+#define CHECK_SIZE(TYPE, CTYPE)						\
+	case TYPE: if (hdr->len < hdr_sz + sizeof(CTYPE)) {		\
+		fprintf(stderr, "config: ERROR "			\
+			#TYPE " message too short for " #CTYPE		\
+			" (%d < %lu)\n",				\
+			hdr->len, hdr_sz + sizeof(CTYPE));		\
+		return 1;						\
+	} else break;
+
+	switch (hdr->type) {
+		CHECK_SIZE(VSOCK_CONFIG_SET_BUF_ALLOC, uint32_t);
+		CHECK_SIZE(VSOCK_CONFIG_SET_CREDIT_LIMIT, uint32_t);
+	default:
+		fprintf(stderr, "config: ERROR "
+			"unknown configuration type %d\n", hdr->type);
+		rc = 1;
+	}
+	return rc;
+#undef CHECK_SIZE
+}
+
 static void apply_socket_config(struct pci_vtsock_sock *s, void *buf)
 {
 	struct vtsock_config_hdr *hdr = (struct vtsock_config_hdr *)buf;
@@ -1712,6 +1738,9 @@ static void apply_socket_config(struct pci_vtsock_sock *s, void *buf)
 			hdr->version);
 		return;
 	}
+
+	if (is_socket_config_malformed(hdr))
+		return;
 
 	switch (hdr->type) {
 	case VSOCK_CONFIG_SET_BUF_ALLOC:
--- a/src/lib/pci_virtio_sock.c
+++ b/src/lib/pci_virtio_sock.c
@@ -1762,7 +1762,7 @@ static int is_socket_config_malformed(struct vtsock_config_hdr *hdr)
 	case TYPE: if (hdr->len < hdr_sz + sizeof(CTYPE)) {		\
 		fprintf(stderr, "config: ERROR "			\
 			#TYPE " message too short for " #CTYPE		\
-			" (%d < %lu)\n",				\
+			" (%d < %zu)\n",				\
 			hdr->len, hdr_sz + sizeof(CTYPE));		\
 		return 1;						\
 	} else break;
@@ -1845,7 +1845,7 @@ static void handle_socket_config(struct pci_vtsock_sock *s, int fd)
 
 	if (sz < sizeof(struct vtsock_config_hdr)) {
 		fprintf(stderr, "config: ERROR "
-			"message length %d is less than header size %lu\n",
+			"message length %d is less than header size %zu\n",
 			sz, sizeof(struct vtsock_config_hdr));
 		return;
 	}
--- a/src/lib/pci_virtio_sock.c
+++ b/src/lib/pci_virtio_sock.c
@@ -1880,7 +1880,8 @@ static ssize_t readrx(struct pci_vtsock_sock *s,
 
 	len = recvmsg(s->fd, &msghdr, 0);
 	if (s->configurable && len > 0 && msghdr.msg_controllen > 0) {
-		fprintf(stderr, "config: received configuration request\n");
+		fprintf(stderr, "config: %s received configuration request\n",
+			s->name);
 		cmsg = CMSG_FIRSTHDR(&msghdr);
 		if (cmsg->cmsg_level == SOL_SOCKET
 		    && cmsg->cmsg_type == SCM_RIGHTS) {
--- a/src/lib/pci_virtio_sock.c
+++ b/src/lib/pci_virtio_sock.c
@@ -140,6 +140,8 @@ struct vtsock_config_hdr {
 #define VSOCK_CONFIG_SET_CREDIT_LIMIT 2
 #define VSOCK_CONFIG_SET_NAME 3
 
+#define MAX_SOCK_NAME_LENGTH 256
+
 struct vsock_addr {
 	uint64_t cid;
 	uint32_t port;
@@ -189,8 +191,6 @@ struct pci_vtsock_sock {
 	 *
 	 *     Allocate DEFAULT_WRITE_BUF_LENGTH memory for ->write_buf
 	 *
-	 *     Allocate and generate a name in ->name
-	 *
 	 *   Drop list_rwlock
 	 *
 	 * To free a sock:
@@ -198,8 +198,6 @@ struct pci_vtsock_sock {
 	 *   Set state to CLOSING_TX and kick the tx thread[*].
 	 *
 	 *   Free ->write_buf
-	 *
-	 *   Free ->name
 	 *
 	 * Then the following will happen:
 	 *
@@ -268,7 +266,7 @@ struct pci_vtsock_sock {
 
 	bool configurable;
 	int id;
-	char *name; /* Used in debugging, allocated with malloc */
+	char name[MAX_SOCK_NAME_LENGTH]; /* Used in debugging */
 
 	struct vsock_addr local_addr;
 	struct vsock_addr peer_addr;
@@ -669,11 +667,7 @@ static struct pci_vtsock_sock *alloc_sock(struct pci_vtsock_softc *sc)
 	s->local_shutdown = 0;
 	s->peer_shutdown = 0;
 
-	if (0 > asprintf(&s->name, "vsock:%d", s->id)) {
-		fprintf(stderr, "Could not allocate memory for vsock name: "
-			"%s\n", strerror(errno));
-		return NULL;
-	}
+	snprintf(s->name, MAX_SOCK_NAME_LENGTH, "vsock:%d", s->id);
 
 	s->write_buf = malloc(s->buf_alloc);
 	if (s->write_buf == NULL) {
@@ -694,7 +688,6 @@ static void free_sock(struct pci_vtsock_softc *sc, struct pci_vtsock_sock *s)
 	LIST_REMOVE(s, list);
 	s->state = SOCK_FREE;
 	free(s->write_buf);
-	free(s->name);
 
 	LIST_INSERT_HEAD(&sc->free_list, s, list);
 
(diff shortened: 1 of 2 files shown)


rneugeba:
I let @ijc25 do the review as this is mostly changes to his code, but it might be worth running `hvgostress` against it. @ijc25 recently extended it to work with vsock as well. You can find it [here](https://github.com/rneugeba/virtsock). Just typing `make` should build it and `examples/README.md` contains some instructions on how to run it on a Mac. the help message as a bunch of options. I would run this for a large number of connections with say `10` or so concurrent connections and maybe larger buffers. Basically let it run for as many hours as you feel like

src/lib/pci_virtio_sock.c
```diff
@@ -618,6 +669,18 @@ static struct pci_vtsock_sock *alloc_sock(struct pci_vtsock_softc *sc)
...
+		fprintf(stderr, "Could not allocate memory for vsock name: "
+			"%s\n", strerror(errno));
+		return NULL;
+	}
+
+	s->write_buf = malloc(s->buf_alloc);
+	if (s->write_buf == NULL) {
+		fprintf(stderr, "Could not allocate memory for write buf: "
+			"%s\n", strerror(errno));
+		return NULL;
```
avsm: need to also `free (s->name)` on this error path

src/lib/pci_virtio_sock.c
```diff
@@ -1626,6 +1711,230 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
...
+	int fd;
+
+	len = recvmsg(s->fd, &msghdr, 0);
+	if (s->configurable && len > 0 && msghdr.msg_controllen > 0) {
+		fprintf(stderr, "config: %s received configuration request\n",
+			s->name);
+		cmsg = CMSG_FIRSTHDR(&msghdr);
+		if (cmsg->cmsg_level == SOL_SOCKET
+		    && cmsg->cmsg_type == SCM_RIGHTS) {
+			memcpy(&fd, CMSG_DATA(cmsg), sizeof(int));
```
yallop: simpler: `fd = CMSG_DATA(cmsg)`

yallop: Indeed -- good point.

src/lib/pci_virtio_sock.c
```diff
@@ -434,6 +441,25 @@ static size_t iovec_clip(struct iovec **iov, int *iov_len, size_t bytes)
 	return ret;
 }
 
+static size_t iovec_shift_left1(struct iovec **iov, int *iov_len)
```
ijc25: A comment explaining what this is supposed do would be appreciated.

Is it actually just the same as `if (iov_len > 0) iovec_pull(&iov, &iov_len, NULL, iov[0].iov_len)`?

src/lib/pci_virtio_sock.c
```diff
@@ -1644,6 +1672,141 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
...
+				strerror(errno));
+			goto err;
+		}
+		if (len == 0) {
+			fprintf(stderr, "config: ERROR "
+				"EOF during config message body read\n");
+			goto err;
+		}
+		read_upto += len;
+	} while (read_upto < sz);
```
ijc25: This is the second appearance of a very similar read loop, refactoring into a `read_exactly` type function would be better.  Sadly I don't think hyperkit has one already.

src/lib/pci_virtio_sock.c
```diff
@@ -1644,6 +1672,141 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
...
+
+	apply_socket_config(s, buf);
+
+	free(buf);
+	goto next_message;
```
ijc25: An infinit loop (`for(;;)` or `while(1)`) would be preferable to goto.

src/lib/pci_virtio_sock.c
```diff
@@ -1644,6 +1672,141 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
...
+/*
+ * Check for an attached socket configuration fd, s->mtx must be held
+ *
+ * If a configuration fd is attached, this will block until the
+ * configuration request is handled!
+ *
+ * Returns > 0 for regular messages, 0 for EOF, -1 for error, and -2
+ * for configuration messages.
+ */
+static ssize_t readrx(struct pci_vtsock_sock *s,
```
ijc25: `readrx` for `read`, `rx`? Perhaps `read_with_control` or `read_with_sideband` or `..._with_config`?

src/lib/pci_virtio_sock.c
```diff
@@ -1644,6 +1672,141 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
...
+		.msg_iov = iov,
+		.msg_iovlen = iov_len,
+		.msg_control = control,
+		.msg_controllen = CMSG_SPACE(sizeof(int /* fd */)),
+	};
+	struct cmsghdr *cmsg;
+	int fd;
+
+	len = recvmsg(s->fd, &msghdr, 0);
+	if (s->configurable && len > 0 && msghdr.msg_controllen > 0) {
```
ijc25: Should it be an error for `msghdr.msg_controllen > 0` to be true when `!s->configurable`?

Is it necessary for `len > 0` to be true in order to have received control data? http://pubs.opengroup.org/onlinepubs/9699919799/functions/recvmsg.html says that `len == 0` is shutdown with no caveats so I guess it must.

src/lib/pci_virtio_sock.c
```diff
@@ -1644,6 +1672,141 @@ static void pci_vtsock_notify_tx(void *vsc, struct vqueue_info *vq)
...
+		fprintf(stderr, "config: received configuration request\n");
+		cmsg = CMSG_FIRSTHDR(&msghdr);
+		if (cmsg->cmsg_level == SOL_SOCKET
+		    && cmsg->cmsg_type == SCM_RIGHTS) {
+			memcpy(&fd, CMSG_DATA(cmsg), sizeof(int));
+			handle_socket_config(s, fd);
+			close(fd);
+			if (len > 1) {
+				iovec_shift_left1(&iov, &iov_len);
```
ijc25: I don't get this at all, what is the relationship between having received a control message and ditching some of the actual data buffer?

dsheets: At least one byte must be transmitted to transmit the fd. The protocol is that the first byte will be ignored if there is an attached fd. I can put this in an informative comment.

src/lib/pci_virtio_sock.c
```diff
@@ -263,6 +264,7 @@ struct pci_vtsock_sock {
 
 	uint32_t buf_alloc;
 	uint32_t fwd_cnt;
+	uint32_t credit_limit;
```
ijc25: This is intentionally unused (apart from to set it) at this point in the series?

src/lib/pci_virtio_sock.c
dsheets: It could be an error or it could continue to be ignored as it is currently. I'm happy to make it an error as it's a strong sign that You're Doing It Wrong.

src/lib/pci_virtio_sock.c
dsheets: Yes.

src/lib/pci_virtio_sock.c
```diff
@@ -652,6 +666,12 @@ static struct pci_vtsock_sock *alloc_sock(struct pci_vtsock_softc *sc)
 	s->local_shutdown = 0;
 	s->peer_shutdown = 0;
 
+	if (0 > asprintf(&s->name, "vsock.%d", id)) {
```
ijc25: `asprintf(...) < 0` is more idiomatic these days.

Up til now I had been using a combination of the src/dst address pairs and the local fd# for identifying debug messages (which is, TBH, rather verbose and not ideal) but how is someone debugging this code supposed map from the id assigned here to an actual flow of interest?

dsheets: Ideally, we would fixup the existing debugging messages to use this name. I can do some of that if you'd like but decided not to go too far down that path as it wasn't necessary (but names were useful in figuring out which mount socket was being configured).

src/lib/pci_virtio_sock.c
```diff
@@ -1723,6 +1724,31 @@ static void apply_socket_config_set_buf_alloc(struct pci_vtsock_sock *s,
...
+static void apply_socket_config_set_name(struct pci_vtsock_sock *s,
+					 struct vtsock_config_hdr *hdr)
+{
+	char *new_name;
+	char *msg_name = (char *)(hdr + 1);
+	size_t name_len = hdr->len - sizeof(struct vtsock_config_hdr);
+
+	name_len = strnlen(msg_name, name_len);
+
+	new_name = malloc(name_len + 22); /* max int string length + : + \0 */
```
ijc25: I'd put the : and \0 inside ' i.e. `+ ':' + '\0'` to mimic C character syntax and separate the characters from the + operators

I would also write `22` as `20 + 1 + 1` to make the relationship to the comment clearer. .

Then that would make me ask for a *#*define for the number 20, wherever that comes from.

Which finally leads me to ask why `s->name` then isn't a static inline array, since it is small and that avoids faffing around with dynamic allocation for this.

ijc25: Oh, I see, 20 is the maximum length of the `%d` formatting of an integer (on this platform at least), which is hard to express as a constant expression. I also see now that this is to be appended to the provided name, hence `s->name` can't be a static array.

Might all this be more clearly expressed as
```
new_thing = asprintf("%*s:%d", msg_len, msg_name, s->id)
free(s->name)
s->name = new_thing
```
??

src/lib/pci_virtio_sock.c
```diff
@@ -1381,6 +1384,7 @@ static void pci_vtsock_proc_tx(struct pci_vtsock_softc *sc,
 			goto do_rst;
 		}
 		vq_relchain(vq, idx, 0);
+		sock->credit_cnt = sock->fwd_cnt - sock->credit_limit;
```
ijc25: A comment indicating that this is the fudge in the commit message would be good, I think.

src/lib/pci_virtio_sock.c
```diff
@@ -2008,6 +2012,7 @@ static ssize_t pci_vtsock_proc_rx(struct pci_vtsock_softc *sc,
 		dprint_header(hdr, 0, "RX");
 		s->credit_update_required = false;
 		vq_relchain(vq, idx, sizeof(*hdr));
+		s->credit_cnt = hdr->fwd_cnt;
```
ijc25: The four places with this `s->credit_update_required = false; s->credit_cnt = hdr->fwd_cnt;` pattern could profitably become `clear_credit_update_required` (cf `set_credit_update_required`) perhaps, which would be somewhat self documenting through the naming.

src/lib/pci_virtio_sock.c
```diff
@@ -1020,12 +1022,13 @@ static void send_response_common(struct pci_vtsock_softc *sc,
...
 static void send_response_sock(struct pci_vtsock_softc *sc,
-				 uint16_t op, uint32_t flags,
-				 const struct pci_vtsock_sock *sock)
+			       uint16_t op, uint32_t flags,
+			       struct pci_vtsock_sock *sock)
 {
 	send_response_common(sc, sock->local_addr, sock->peer_addr,
 			     op, VIRTIO_VSOCK_TYPE_STREAM, flags,
 			     sock->buf_alloc, sock->fwd_cnt);
+	sock->credit_cnt = sock->fwd_cnt;
```
ijc25: Is this the correct place for this? At this point we are just queueing the response onto the ring which goes between our local tx and rx threads, so it won't yet be visible to the driver.

Would it be more correct to defer this update to the point where the rx thread dequeues and processes it? That's the point at which the driver could be said to be about to see this update. That would mean somewhere like in ` pci_vtsock_proc_rx`, which actually it looks like you have as well?

ijc25:
The complete change was a bit hard to review so I went through https://github.com/docker/hyperkit/pull/92/commits in the order presented, which didn't seem to make sense, GH might be ordering things by commit date rather than parentage and your (correct) rebasing has confused it.

I commented on various things as I went for the rest LGTM on:
* vsock: indent with tabs
* vsock: allocate s->buf_alloc on the heap
* vsock: add s->credit_limit field and configuration message to set it
* vsock: add malformed config message check
* vsock: print socket name that receives config request 
* vsock: add a SET_BUF_ALLOC configuration message
* vsock: factor inbound connection request parsing

src/lib/pci_virtio_sock.c
ijc25: What memory is shared here and with whom?

The control byte is something in the protocol you've designed, due to the comment where I asked about len > 0 and the interaction with the control and data parts of the message? (i.e. you need to send some dummy data to get the control message through?) If so that's definitely worth spelling out somewhere.

src/lib/pci_virtio_sock.c
ijc25: I'm inclined towards being picky in what we receive in this case and rejecting it.

src/lib/pci_virtio_sock.c
ijc25: Thanks, I figured this out from other responses to other queries too. A comment would be excellent, thanks.

dsheets:
GitHub indeed seems to think that commit timestamp is the correct dimension for ordering rather than commit order.

I will address your comments in new commit(s) and ping you when done. Then, I can rebase to fixup the original set if you'd like. Thanks for your careful review!

ijc25:
Thinking about the dummy byte required to send a control message overnight, is this pattern on the other end handled correctly:
```
write(N byte buffer);
sendmsg(control with 1 byte dummy)
write(M byte buffer);
```
Does the hyperkit end see:
```
read(N byte buffer);
read(1 byte dummy with control message);
read(M byte buffer);
```
Or does it see
```
read(N+1+M byte buffer with control message);
```
(assuming scheduling is such that all writes happen before hyperkit even gets a look)

IIRC at one point you mooted only accepting control messages before any data had been sent, but it looks like you changed your mind on that since you only clear `s->configurable` upon receipt of a control fd and not for any successful read.

src/lib/pci_virtio_sock.c
dsheets: That would be nicer -- for some reason I forgot about `asprintf` temporarily! As discussed, I think we should probably just truncate long names and have then statically allocated in the sock struct.

src/lib/pci_virtio_sock.c
dsheets: No, this is not the correct place. I was a bit uncertain about this one but I agree with your assessment that it is premature to update `sock->credit_cnt` here.

dsheets:
`s->configurable` should be cleared after the first read. I had incorrectly thought that ancillary data would induce framing but that is incorrect and I never updated this logic to reflect that. Thanks for catching that!

src/lib/pci_virtio_sock.c
```diff
@@ -130,6 +130,11 @@ static int pci_vtsock_debug = 0;
 
 /* XXX need to use rx and tx more consistently */
 
+struct vtsock_config_hdr {
+	uint32_t len;
+	uint32_t version;
+};
```
dsheets: The `type` member was added in a4ec860c5600d43b3103e7e7a9394bcc6a070fd6 but I subsequently rewrote the initial protocol commit to reflect the intent. I can fix this up if you think it would improve clarity.

ijc25: You _could_ move the addition of `type` to the commit which introduces the protocol and `struct vtsock_config_hdr` and include a dummy:
```
	switch (hdr->type) {
	default:
		fprintf(stderr, "config: ERROR "
			"unknown configuration type %d\n", hdr->type);
	}
```
where your eventual switch would be, but it's probably not worth the rebase faff now that you've explained the temporary inconsistency in that one commit here

PR #93, opened 2016-12-10
ocaml: qcow >= 0.7.0 is functorised over Time

djs55:
Modern versions of `qcow` when configured to automatically compact will wait until there have been no `discard`s for some period of time in order to batch a run of `discard`s together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ via `brew` and using `opam` to install the appropriate libraries:
     $ brew install opam libev
     $ opam init
     $ eval `opam config env`
-    $ opam install uri qcow conf-libev logs fmt
+    $ opam install uri qcow.0.7.0 conf-libev logs fmt
 
 Notes:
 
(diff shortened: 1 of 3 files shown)


ijc25:
Does this require a version constraint adding somewhere? I'm not sure where it would be, we don't seem to have an opam file in this repo (nor am I sure we want/need one).

Assuming we don't need anything like that: LGTM.

djs55:
I don't think we strictly need a constraint -- `opam` will install the latest version which is now `0.7.0`.

However when I changed the interface upstream it broke master here, which isn't too good. I think the place a constraint could go would be in the `README.md`: something like 
```
    $ opam install uri qcow.0.7.0 conf-libev logs fmt
```
It's unfortunate that I've changed the interface twice in the last two weeks. Thinking about it, Mirage v3 due in about a month will change the APIs again. I'll modify the `README.md` now while we're thinking about it!

ijc25:
Extra change LGTM, we should remember it when 0.8.0 comes along though ;-)

djs55:
Thanks!

PR #97, opened 2017-01-17
Increase the maximum number of vCPUs to 64

### LINKED ISSUE #96: Allow more than 16 vCPUs
Currently <https://github.com/docker/hyperkit/blob/master/src/include/xhyve/vmm/vmm_common.h#L33> restricts the maximum number of vCPUs to 16. For compute-heavy workloads, this arbitrary restriction may be undesirable (see docker/for-mac#1144). This number is mainly used to size static arrays but, as @ijc25 notes, there may be work checking that ACPI tables are set up correctly to offer enough cores correctly.

rneugeba:
Increase the MADT size accordingly. Add a compile time
check to make sure the MADT doesn't overflow into the FADT.

resolves #96

### DIFF
--- a/src/include/xhyve/vmm/vmm_common.h
+++ b/src/include/xhyve/vmm/vmm_common.h
@@ -30,7 +30,7 @@
 
 #include <stdint.h>
 
-#define	VM_MAXCPU 16 /* maximum virtual cpus */
+#define	VM_MAXCPU 64 /* maximum virtual cpus */
 
 enum vm_suspend_how {
 	VM_SUSPEND_NONE,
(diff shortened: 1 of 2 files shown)


ijc25:
I think the ACPI code should error out if the number of vcpus is enough to overflow the MADT. Peter [suggested he would do the same in bhyve](https://docs.freebsd.org/cgi/getmsg.cgi?fetch=130863+0+/usr/local/www/mailindex/archive/2016/freebsd-virtualization/20160918.freebsd-virtualization) so you might be able to grab a hunk of patch from upstream?

AIUI the current generation of Macs top out at 24 logical cores (2x6 physical core x2 hyperthreads each). It not implausible that the next h/w refresh will bump things past 32 (although given Skylake looks to top out at 8 threads it seems like going to exactly 32 is most plausible) maybe we should just bite the bullet and go to max=64 (with corresponding increase to MADT) while we are at it?

rneugeba:
Updated the patch to go to 64 vCPUs. Increased the MADT, updated comments and added a compile time check for the MADT size.

Also added the PR text/title to reflect the changes

ijc25:
LGTM

PR #98, opened 2017-01-20
Dump guest physical and linear address from VMCS when dumping state

ijc25:
These may or may not be valid for a particular fault but other
information which is already logged (such as the vmexit qualification)
should allow a developer to work out if the values are valid on a case
by case basis.


rneugeba:
LGTM

PR #99, opened 2017-02-13
Update to Mirage 3

djs55:
This PR updates to the new Mirage v3 block API and includes the latest version of the `qcow` library which supports flushing via `fsync` as well as `fcntl`: see [docker/for-mac#668]

Note the Mirage 3 packages have been tagged and released but the package metadata is in a new repository: https://github.com/mirage/mirageos-3-beta. This PR updates the `circle.yml` and the instructions in the `README.md` accordingly.

### DIFF
--- a/src/lib/mirage_block_ocaml.ml
+++ b/src/lib/mirage_block_ocaml.ml
@@ -8,11 +8,7 @@ module Log = (val Logs.src_log src : Logs.LOG)
 
 
 (* TODO: parameterise this over block implementations *)
-module Time = struct
-  type 'a io = 'a Lwt.t
-  let sleep_ns ns = Lwt_unix.sleep (Int64.to_float ns /. 1e9)
-end
-module Qcow = Qcow.Make(Block)(Time)
+module Qcow = Qcow.Make(Block)(OS.Time)
 
 module Mutex = struct
   include Mutex
--- a/src/lib/mirage_block_ocaml.ml
+++ b/src/lib/mirage_block_ocaml.ml
@@ -10,7 +10,7 @@ module Log = (val Logs.src_log src : Logs.LOG)
 (* TODO: parameterise this over block implementations *)
 module Time = struct
   type 'a io = 'a Lwt.t
-  let sleep = Lwt_unix.sleep
+  let sleep_ns ns = Lwt_unix.sleep (Int64.to_float ns /. 1e9)
 end
 module Qcow = Qcow.Make(Block)(Time)
 
@@ -88,7 +88,7 @@ module Protocol = struct
       | Write of int
       | Delete
       | Flush
-    type t = [ `Ok of ok | `Error of Qcow.error ]
+    type t = (ok, Qcow.write_error) result
   end
 
   (* An in-flight request *)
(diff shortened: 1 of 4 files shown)


ijc25:
LGTM (although I'm not really qualified to review all the ocaml changes, maybe ping a friendly ocaml person for a quick sanity check?)

Am I right that eventually the mirage-3 stuff will be fully released and appear in a mainline repository (i.e. not mirage-3-beta)? If so could you please keep an eye out for that event and update this stuff accordingly?

djs55:
Thanks -- @samoht could I borrow your eyes for a quick sanity-check?

You're correct that the Mirage stuff should be fully released soonish.. at that point I'll make another PR to tidy this up a bit.

src/lib/mirage_block_ocaml.ml
```diff
@@ -10,7 +10,7 @@ module Log = (val Logs.src_log src : Logs.LOG)
 (* TODO: parameterise this over block implementations *)
 module Time = struct
   type 'a io = 'a Lwt.t
-  let sleep = Lwt_unix.sleep
+  let sleep_ns ns = Lwt_unix.sleep (Int64.to_float ns /. 1e9)
```
samoht: You will probably be interested by https://hannesm.github.io/duration/doc/Duration.html#VALof_sec

src/lib/mirage_block_ocaml.ml
```diff
@@ -285,66 +285,77 @@ open Lwt
...
-            let len = List.(fold_left (+) 0 (map Cstruct.len bufs)) in
-            return (Response.Read len)
+            let sector = Int64.of_int (offset / t.Handle.info.Mirage_block.sector_size) in
+            let open Lwt.Infix in
+            (* Qcow.error <> Qcow.write_error *)
+            begin Qcow.read t.Handle.block sector bufs
+              >>= function
+              | Error `Disconnected -> Lwt.return (Error `Disconnected)
+              | Error `Unimplemented -> Lwt.return (Error `Unimplemented)
+              | Ok () ->
```
samoht: You can probably shorten that match to:
```
| Error _ as e -> Lwt.return e
| Ok () -> ...
```

src/lib/mirage_block_ocaml.ml
```diff
@@ -360,16 +371,23 @@ let process_one t =
...
             let t = Handle.find_or_quit h in
-            Block.flush t.Handle.base
-            >>= fun () ->
-            return Response.Flush
+            let open Lwt.Infix in
+            (* Block.write_error <> Qcow.write_error *)
+            begin Block.flush t.Handle.base
+              >>= function
+              | Error `Disconnected -> Lwt.return (Error `Disconnected)
+              | Error `Is_read_only -> Lwt.return (Error `Is_read_only)
```
samoht: Same here, you can write this as:

```
| Error _ as e -> Lwt.return e
| Ok () -> return Response.Flush
```

and you can also probably remove the `begin ... end`

djs55: I'm probably doing something wrong (perhaps elsewhere, in `ocaml-qcow`?) but I get
```
File "src/lib/mirage_block_ocaml.ml", line 376, characters 12-435:
Error: This expression has type
         (Protocol.Response.ok, Block.write_error) result Lwt.t
       but an expression was expected of type
         (Protocol.Response.ok, Qcow.write_error) result Lwt.t
       Type
         Block.write_error =
           [> `Disconnected | `Is_read_only | `Unimplemented ]
       is not compatible with type
         Qcow.write_error =
           [> `Disconnected | `Is_read_only | `Unimplemented ] 
make: *** [build/lib/mirage_block_ocaml.o] Error 2
```

djs55: Sorry I deleted my comment because I thought I'd typo'd something. I've double-checked and it still doesn't work for me, which is probably a sign of a problem in `ocaml-qcow`?
```
File "src/lib/mirage_block_ocaml.ml", line 376, characters 12-432:
Error: This expression has type
         (Protocol.Response.ok, Block.write_error) result Lwt.t
       but an expression was expected of type
         (Protocol.Response.ok, Qcow.write_error) result Lwt.t
       Type
         Block.write_error =
           [> `Disconnected | `Is_read_only | `Unimplemented ]
       is not compatible with type
         Qcow.write_error =
           [> `Disconnected | `Is_read_only | `Unimplemented ] 
```
WDYT?

samoht: Hum, can you try `| Error e -> Lwt.return (Error e)` maybe?

samoht: But yes, the type of error should probably be made explicit in https://github.com/mirage/ocaml-qcow/blob/master/lib/qcow.mli#L21 (e.g. using `with type write_error := [ Block.write_error | `Msg of string]`

djs55:
Thanks for the feedback! I just realised that there's already an implementation of the time signature in [mirage-unix](https://github.com/mirage/mirage-platform/blob/master/unix/lib/time.mli) which I could re-use rather than re-implement.

Regarding the `Error e` matches: after speaking to @samoht in detail I think I understand the problem is with the current version of `ocaml-qcow`. Since this is released already I'll merge this as-is, and when I fix the qcow `write_error` problem and re-release I'll circle back here.

PR #104, opened 2017-03-16
go: Add a simple go wrapper around the HyperKit executable

rneugeba:
The wrapper comes as a Go package which can be used by
other code to create/start/stop hyperkit VMs as well as
a small utility for testing purposes.

The functionality is currently limited to creating basic
Linux based VMs, i.e., some of the options are hardcoded.


go/hyperkit.go
```diff
@@ -0,0 +1,503 @@
...
+
+	done := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case stderrl := <-stderrChan:
+				fmt.Println("HyperKit STDERR: ", stderrl)
+			case stdoutl := <-stdoutChan:
+				fmt.Println("HyperKit STDOUT: ", stdoutl)
+			case <-done:
```
ijc25: You might as well just hook them up to stdio if you do this. Will probably do for now though.

go/hyperkit.go
```diff
@@ -0,0 +1,503 @@
...
+	cmd := exec.Command(h.HyperKit, h.Arguments...)
+	cmd.Env = os.Environ()
+
+	if h.Console == ConsoleStdio {
+		// Plumb in stdin/stdout and run in foreground
+		cmd.Stdin = os.Stdin
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		h.writeState()
+		err := cmd.Run()
```
ijc25: This blocks until hyperkit exits, which might not be expected, especially given it doesn't block in the `h.Console != ConsoleStdio` case so is inconsistent based on a slightly non-obvious condition, I know it is documented but it still violates the principal of least surprise.

The public interface is `Start` which has implications from `os/exec:Start` that it will run in the background. Calling `Start` here instead would also solve the thing where you have to jump through hoops to get the `Pid`, maybe you would then also want to expose the underlying `Wait` operation too. In fact you could consider exposing the full `Run` vs `Start+Wait` semantics to the caller and let them make the choice directly rather than indirectly via the console mechanism they've chosen.

go/Dockerfile.build
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.8-alpine
+RUN apk update && apk add --no-cache make git
+ENV GOPATH=/go
+ENV PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+# The project sources
+VOLUME ["/go/src/github.com/docker/hyperkit/go"]
+WORKDIR /go/src/github.com/docker/hyperkit/go
+
+ENTRYPOINT ["make"]
```
ijc25: Is a Linux build of this functionality useful given it can't fork+exec hyperkit in its present form?

ijc25:
Would be good to hook up at least a build (ideal on Darwin) in the `circle.yml`, if not an actual runtime test too.

go/Dockerfile.build
rneugeba: I tend to cross compile Go code and GOOS and GOARCH are set to OS X in the Makefile. Compiling in a container this way frees me from having to stick all my repositories into GOPATH

ijc25: I missed the `GOOS=darwin` in the `Makefile`. I think we also do need to support building without docker (i.e. with things in the correct places in `$GOPATH`). Reading the `Makefile` it looks like that should work if you call the correct `Makefile` target.

rneugeba: yes. will call it when I try circle-ci integration. Need to checkout the repo at correct place for that, though

go/hyperkit.go
rneugeba: good point. Thinking of doing:
- Remove `Console` option
- `hyperkit.Start()`: starts in the background, requires `StateDir` to be set to stash info
- `hyperkit.Run()`: starts in foreground with stdio plumbed through

This allows for the two common use cases, I'm interested in. Starting in the foreground with `tty` file and `console-ring` would not be supported. @ijc25 do you think this would be an issue?

ijc25: Strictly speaking the hyperkit stdio is still valid/relevant when the console goes elsewhere, it doesn't have the emulated UART mapped to it but it still gets hypervisor logging and error messages.

You could expose the underlying `Std???Pipe` methods too and then it would work (if the caller wanted it to)? You are right that it is probably a less important usecase though.

rneugeba:
Force pushed an update:
- We now have `Start()` and `Run()` so the caller explicitly asks for background/foreground execution.
- The `hyperkit` process stdout/stderr are written to a logger if configured
- CircleCI builds the sample binary and runs gofmt and golint for checks
- renamed `cmd` to `sample` as this really is just sample code.

circle.yml
```diff
@@ -5,6 +5,10 @@ general:
 machine:
   xcode:
     version: "7.3"
+  environment:
+    GOPATH: "$HOME/go"
+    PROJECT_ROOT: "$GOPATH/src/github.com/docker/hyperkit"
+    PATH: "${PATH}:${GOROOT}/bin:${GOPATH}/bin"
```
ijc25: Is `${GOROOT}` not in some default location? Does circleci provide `${GOROOT} for us?

rneugeba: It appears to be set. Come to think of it, I think setting the the `PATH` based on `GOPATH` should be sufficient. This is for it to find `golint` which should be installed in `GOPATH`

ijc25: Yes, I've never knowingly set `GOROOT` or used it for anything. Lets try without!

circle.yml
```diff
@@ -18,3 +22,8 @@ test:
     - eval `opam config env` && make all
     - make test
     - eval `opam config env` && make test-qcow
+    # Go related steps
+    - mkdir -p $(dirname "$PROJECT_ROOT")
+    - ln -s $(pwd) "$PROJECT_ROOT"
+    - go get github.com/golang/lint/golint
```
ijc25: Move it into the Makefile in a `setup` rule perhaps? More convenient for non-CI use there.

rneugeba:
Removed the `GOROOT` based path and move the `golint` install to Makefile. CI is still green

go/hyperkit.go
```diff
@@ -0,0 +1,524 @@
...
+
+	return nil
+}
+
+// Stop the running VM
+func (h *HyperKit) Stop() error {
+	if !h.IsRunning() {
+		return nil
+	}
+	p, err := os.FindProcess(h.Pid)
```
ijc25: In the common case can't you remember the `cmd.Process` which you launched which will be more reliable than trying to figure it out here? And in the uncommon case where you reattach from state would it not be better to do this in `FromState`?

rneugeba: Yes, this makes it cleaner.

Note, currently (and with the new version which I'll push soon) the Stop() and IsRunning() only work on background processes. For foreground processes we'd have to stash the `exec.Cmd` reference somewhere, poke around in there to find the process etc.

ijc25: Don't you already stash the corresponding `os.Process` thing in your hyperkit object?

rneugeba:
I think you looked at an outdated version. It now uses:
```
// Stop the running VM
func (h *HyperKit) Stop() error {
        if h.process == nil {
                return fmt.Errorf("hyperkit process not known")
        }
        if !h.IsRunning() {
                return nil
        }
        err := h.process.Kill()
        if err != nil {
                return err
        }

        return nil
}
```

go/hyperkit.go
```diff
@@ -0,0 +1,528 @@
...
+func CreateDiskImage(location string, sizeMB int) error {
+	f, err := os.Create(location)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	buf := make([]byte, 1048676)
+	for i := 0; i < len(buf); i++ {
+		buf[i] = 0
```
ijc25: https://blog.golang.org/go-slices-usage-and-internals says:
> Arrays do not need to be initialized explicitly; the zero value of an array is a ready-to-use array whose elements are themselves zeroed: 

This is technically a slice, but there must be an underlying array so I think this holds and the init is unnecessary.

http://stackoverflow.com/questions/16797380/how-to-create-a-10mb-file-filled-with-000000-data-in-golang suggests `f.Truncate(size)` might be workable, not sure about on OSX due to lack of sparse files but might save this entire wodge of code.

go/hyperkit.go
```diff
@@ -0,0 +1,528 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// Convert to json string
+func (h *HyperKit) String() string {
+	s, _ := json.Marshal(h)
+	return string(s)
```
ijc25: `if err != nil return string(err)` perhaps (in proper syntax). Better than getting an empty result?

go/hyperkit.go
```diff
@@ -0,0 +1,528 @@
...
+		}
+		stream(stdout, stdoutChan)
+		stream(stderr, stderrChan)
+
+		done := make(chan struct{})
+		go func() {
+			for {
+				select {
+				case stderrl := <-stderrChan:
+					if h.log != nil {
```
ijc25: You should hoist this h.log up to the top level, if it is nil then route stdio direct to `/dev/null` and only construct the channels and start the goroutine pump if it is non-nil.

rneugeba:
Thanks for the good suggestion. Addressed in a new force push

go/hyperkit.go
```diff
@@ -0,0 +1,526 @@
...
+	return nil
+}
+
+// IsRunning returns true if the hyperkit process is running.
+func (h *HyperKit) IsRunning() bool {
+	// os.FindProcess on Unix always returns a process object even
+	// if the process does not exists. There does not seem to be
+	// a call to find out if the process is running either, so we
+	// use another package to find out.
+	proc, err := ps.FindProcess(h.Pid)
```
ijc25: Shouldn't this function use `h.process` like `Stop` et al? (especially since `Stop` uses this function internally too).

ijc25: Oh wait, this handles the case where it has died since you last looked at it. You can't use `Wait` as a trigger in the case where you have attached to an existing process, so this is as good as it gets...

go/hyperkit.go
```diff
@@ -0,0 +1,526 @@
...
+		err := cmd.Start()
+		if err != nil {
+			return err
+		}
+		h.Pid = cmd.Process.Pid
+		h.process = cmd.Process
+		h.writeState()
+	} else {
+		h.writeState()
+		err := cmd.Run()
```
ijc25: You could make this:
```
cmd.Start()
Fill in h.PId and h.Process
call h.writeState()
cmd.Wait()
```
I think `Start() + Wait()` is semantically the same as `Run()`, but with the opportunity to do stuff in the middle, isn't it?

rneugeba:
checked and `Run()` indeed is the same as `Start`/`Wait`. So that simplified the code further.

go/hyperkit.go
```diff
@@ -0,0 +1,524 @@
...
+		return err
+	}
+	h.Pid = cmd.Process.Pid
+	h.process = cmd.Process
+	h.writeState()
+	if !h.background {
+		err = cmd.Wait()
+		if err != nil {
+			return err
+		}
```
ijc25: idomatic go would be
```
if err := cmd.Wait(); err != nil {
   return err
}
```
Here and in various other places where you don't need one of the other results from the function call, but that's down to your personal preference as author so don't feel obliged to change...

go/hyperkit.go
```diff
@@ -0,0 +1,524 @@
...
+	return nil
+}
+
+// writeState write the state to a JSON file
+func (h *HyperKit) writeState() {
+	if h.StateDir == "" {
+		return
+	}
+
+	s, _ := json.Marshal(h)
```
ijc25: You should probably propagate this error, or else other things might get quite confused trying to unpickle it later.

go/hyperkit.go
rneugeba: personally, I don't like the idiomatic way and since this is not picked up by lint...

rneugeba:
update pushed

go/hyperkit.go
```diff
@@ -0,0 +1,531 @@
...
+		return nil, fmt.Errorf("Can't read json file: ", err)
+	}
+	h := &HyperKit{}
+	err = json.Unmarshal(b, h)
+	if err != nil {
+		return nil, fmt.Errorf("Can't parse json file: ", err)
+	}
+
+	// We can't get the pid into the JSON if we run it in the
+	// foreground so try reading it from the pid file.
```
ijc25: No longer true?

ijc25: Although their may be mileage in confirming the pid file and the json agree...

go/hyperkit.go
```diff
@@ -0,0 +1,531 @@
...
+		stream(stderr, stderrChan)
+
+		done := make(chan struct{})
+		go func() {
+			for {
+				select {
+				case stderrl := <-stderrChan:
+					log.Printf("Hyperkit ERR: ", stderrl)
+				case stdoutl := <-stdoutChan:
+					log.Printf("Hyperkit OUT: ", stdoutl)
```
ijc25: Assume that the logger will prefix it desired?

go/hyperkit.go
```diff
@@ -0,0 +1,531 @@
...
+	}
+	return ioutil.WriteFile(filepath.Join(h.StateDir, jsonFile), []byte(s), 0644)
+}
+
+func stream(r io.ReadCloser, dest chan<- string) {
+	go func() {
+		defer r.Close()
+		reader := bufio.NewReader(r)
+		for {
+			line, err := reader.ReadString('\n')
```
ijc25: I wonder what happens if the number of chars before a `\n` is bigger than the internal buffer of the `bufio.Reader`? I think this isn't exposed to the actual console (just the hyperkit process' own output) so can probably ignore that concern.

rneugeba:
nice spot re the PID. pushed update. Also removed prefix logger

go/hyperkit.go
```diff
@@ -0,0 +1,532 @@
...
+		stream(stderr, stderrChan)
+
+		done := make(chan struct{})
+		go func() {
+			for {
+				select {
+				case stderrl := <-stderrChan:
+					log.Printf(stderrl)
+				case stdoutl := <-stdoutChan:
+					log.Printf(stdoutl)
```
ijc25: In the absence of a `log.Puts` this would be better as `Printf("%s", stdoutl)`, in case of % chars in the stream. (same for err)

rneugeba:
good call with the "%" in the output streams. Update

go/hyperkit.go
```diff
@@ -0,0 +1,532 @@
...
+
+	err := cmd.Start()
+	if err != nil {
+		return err
+	}
+	h.Pid = cmd.Process.Pid
+	h.process = cmd.Process
+	err = h.writeState()
+	if err != nil {
+		return err
```
ijc25: Should this kill the process?

ijc25:
I made one more comment but TBH lets just merge this sucker and fixup any remaining issues as and when...

go/hyperkit.go
rneugeba: yes. pushed an update

rneugeba:
Thanks for the detailed review!

PR #111, opened 2017-03-28
Change the exit codes to something saner

### LINKED ISSUE #108: hyperkit returns 1 on clean exit/powerdown
Running `hyperkitrun.sh` and use `sudo poweroff` cause the hyperkit process to exit with code 1. It would be nice if it exits with 0 on a clean guest shutdown

rneugeba:
- Use 0 for both poweroff and halt. They are more or less the same.
- Use 2 when the guest rebooted. This allows the user to restart
  the process, possibly with some changes, such as removing a
  CDROM device.

The other exit codes are untouched.

Also document the exit codes in the man page while at it.

resolves #108

### DIFF
--- a/hyperkit.1
+++ b/hyperkit.1
@@ -356,6 +356,24 @@ hyperkit -c 4 \\
 
 .Ed
 
+.Sh EXIT STATUS
+The
+.Nm hyperkit
+process exits with one of the following values:
+.Pp
+.Bl -tag -width flag -compact
+.It Li 0
+The guest either powered down or halted cleanly.
+.It Li 1
+An error occurred.
+.It Li 2
+The guest rebooted/reset. The caller is responsible for starting a new hyperkit process in this case.
+.It Li 3
+A triple fault occurred.
+.It Li 100
+An invalid/unknown VM exit occurred.
+.El
+
 .Sh SIGNALS
 .Bl -tag -width 10n
 .It Pa SIGUSR1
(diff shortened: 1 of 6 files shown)


hyperkit.1
```diff
@@ -356,6 +356,24 @@ hyperkit -c 4 \\
...
+.Nm hyperkit
+process exits with one of the following values:
+.Pp
+.Bl -tag -width flag -compact
+.It Li 0
+The guest either powered down or halted cleanly.
+.It Li 1
+An error occurred.
+.It Li 2
+The guest rebooted/reset. The process may needs restarting.
```
ijc25: `s/needs/need/`.

I think there is also no "may" about it -- if the caller wants reboot to mean something then it must restart. How about "The guest rebooted/reset. The caller is responsible for starting a new hyperkit process in this case"?

rneugeba: I like your wording. Force pushed it.

rneugeba:
Might as well make turn the `exit(-1)` into `exit(1)` for consistency. The `exit(0)` in `pci_vtnet_init()` seems alright. I think it's there for you to query the MAC address...

Anyway, pushed an update with the above changes

ijc25:
> The exit(0) in pci_vtnet_init() seems alright. I think it's there for you to query the MAC address...

Yeah, I guess I don't really see why that is useful, especially given that it configures the whole hypervisor just to exit early with that information. It adds an option to the CLI which only makes sense with one particular `-s` configuration too. Lastly I'm not sure what guarantee there is you will get the same MAC on your next attempt (i.e. if you were using this to probe before or after launching hyperkit). Anyway, not really anything to do with this PR so lets ignore...

LGTM.

PR #112, opened 2017-03-28
Remove "-M"/"print_mac" option.

rneugeba:
This global option was specific to one network backend,
irrespective of it being configured via -s.

It is also uncertain that a VM will get a the same MAC
address on a second invocation, so the usefulness of the
option is questionable.


ijc25:
It was also improperly indented, which is its biggest sin IMHO :-)

LGTM

rneugeba:
Indeed...and undocumented

PR #113, opened 2017-03-28
Also remove the -M option from the help message

rneugeba:
Missed this in the previous commit.


ijc25:
LGTM

PR #114, opened 2017-03-28
Some improvements to the go bindings

### LINKED ISSUE #109: Expose `virtio-sock` in go bindings
Some downstream projects may want to use `virtio-sock` via the go bindings. Need to expose this in the API.

rneugeba:
- optionally enable the virtio-socket device (resolves #109)
- add a option to sample code to enable the virtio-socket device
- allow the hyperkit package to be used on non-macOS systems (to make it easier for downstream users to consume)

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -73,6 +73,8 @@ type HyperKit struct {
 	DiskImage string `json:"disk"`
 	// ISOImage is the (optional) path to a ISO image to attach
 	ISOImage string `json:"iso"`
+	// VSock enables the virtio-socket device and exposes it on the host
+	VSock bool `json:"vsock"`
 
 	// Kernel is the path to the kernel image to boot
 	Kernel string `json:"kernel"`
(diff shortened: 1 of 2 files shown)


go/hyperkit.go
```diff
@@ -355,6 +358,9 @@ func (h *HyperKit) buildArgs(cmdline string) {
 	if h.DiskImage != "" {
 		a = append(a, "-s", fmt.Sprintf("2:0,virtio-blk,%s", h.DiskImage))
 	}
+	if h.VSock {
+		a = append(a, "-s", fmt.Sprintf("3,virtio-sock,guest_cid=3,path=%s", h.StateDir))
```
ijc25: You may want to consider appending a subdir to `h.StateDir` so the various Unix sockets are segregated from the rest of the state stuff and avoid potential conflicts.

Otherwise this all LGTM

rneugeba: On Docker for Mac we mostly keep them in the same directory. I leave them as is for now and re-arrange if it becomes an issue

PR #115, opened 2017-04-03
go: Fix some lint complains

rneugeba:
The three Errorf lines where called out by one downstream project.
Some versions of lint seem pickier than others ...


ijc25:
Error's should be formatted as `%s` which DTRT, using `%v` prints opaque internal pointers and the like.

rneugeba:
Ah, I think I just copied this from somewhere else...fixed now

ijc25:
LGTM once CI says it is ok.

PR #116, opened 2017-04-05
go: Expose the VPNKit UUID via the go binding

rneugeba:
The UUID allows a VM to get more consistent network settings.


ijc25:
LGTM

PR #117, opened 2017-04-07
go: Make sure we reap child hyperkit processes if backgrounded

rneugeba:
In some use cases, the go bindings may be used to start
hyperkit processes in the background and the processes may
get killed externally. reaping the child process makes
sure that there are no zombie processes.

An alternative would be to double fork(), but that seems a bit
iffy in Go.


go/hyperkit.go
```diff
@@ -544,3 +548,14 @@ func getHome() string {
...
+// from http://stackoverflow.com/questions/24153268/how-do-i-write-an-async-child-reaper-in-go
+func reapChild(cmd *exec.Cmd) {
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGCHLD)
+	go func() {
+		_ = <-sigc
+		cmd.Wait()
+		signal.Stop(sigc)
+	}()
+}
```
ijc25: Can't you just do `go cmd.Wait()` at the top level (where you call `reapChild` here) to achieve the same thing?

If the process has multiple children and any one of them other than `cmd` exits then this code basically degrades into that anyway.

Otherwise you need a for loop and `wait4` with `WNOHANG` to keep handling `SIGCHLD` and doing a non-blocking check if it was the one you were interested in until it is.

rneugeba: good point, will try the wait in a go routine

rneugeba:
`go cmd.Wait()` seems to do the job

ijc25:
LGTM.

PR #118, opened 2017-04-10
Simplify the go package a little

rneugeba:
- Remove the UserData option. Users can roll their own ISO and pass that in.
- Don't treat the diskimage option special. Configure a disk, just like the other options
- Create the path to the diskimage if it doesn't exit


ijc25:
LGTM

PR #120, opened 2017-04-12
[WIP] Change meaning of UUID field, add VPNKitKey field

errordeveloper:
We have missused UUID filed for VPNKit, and not the `-U` flag which it naturally maps to.

It seems reasonably to leave it up to the caller how they wish to generate UUIDs.

This is a breaking change, hence I marked it WIP (so it doesn't gets merged too soon), and the biggest question question is how should we communicate this change to those who depend on these bindings?

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -69,7 +69,9 @@ type HyperKit struct {
 	StateDir string `json:"state_dir"`
 	// VPNKitSock is the location of the VPNKit socket used for networking.
 	VPNKitSock string `json:"vpnkit_sock"`
-	// UUID is a string containing a UUID for the VM. It can be used in conjunction with VPNKit to get consistent IP address.
+	// VPNKitKey is a string containing a UUID, it can be used in conjunction with VPNKit to get consistent IP address.
+	VPNKitKey string `json:"vpnkit_key"`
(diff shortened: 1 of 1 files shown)


go/hyperkit.go
```diff
@@ -69,7 +69,9 @@ type HyperKit struct {
 	StateDir string `json:"state_dir"`
 	// VPNKitSock is the location of the VPNKit socket used for networking.
 	VPNKitSock string `json:"vpnkit_sock"`
-	// UUID is a string containing a UUID for the VM. It can be used in conjunction with VPNKit to get consistent IP address.
+	// VPNKitKey is a string containing a UUID, it can be used in conjunction with VPNKit to get consistent IP address.
+	VPNKitKey string `json:"uuid"`
```
ijc25: json should be `vpnkit_key` I think. No idea what happens with the duplicated names!

errordeveloper: Thanks, I've missed that!

errordeveloper:
It seems reasonably to leave it up to the caller how they wish to generate UUIDs.

ijc25:
LGTM, these bindings are pretty new so no need to worry about making breaking just yet.

In the future it might be nice to generate a UUID if the caller doesn't supply one.

PR #124, opened 2017-04-20
Generate LICENSE file

guillaumerose:
Projects using Hyperkit need its license. This is the code that generate it in Docker for Mac project. 

I propose to reuse it here, or perhaps we should use a simple LICENSE file at root of the repository ?


djs55:
This looks good to me. If someone wants to build (and maintain) a centralised `LICENSE` file in future they could replace this logic.

PR #125, opened 2017-04-25
Circleci artifacts work on macOS 10.10

guillaumerose:
Circleci binary was not compatible with macOS 10.10. In Docker for Mac, we need to support this version (at least for some weeks).


ijc25:
LGTM

PR #126, opened 2017-04-27
Do not require trailing commas for firmware definition

pborzenkov:


hyperkit.1
```diff
@@ -275,18 +275,18 @@ Disable MPtable generation.
...
+.It fbsd , Ns Pa userboot , Ns Pa bootvolume Ns Oo , Ns Pa kernelenv Ns Oc
 Boot using the fbsd protocol.
-.It bootrom , Ns Pa path , Ns ,
+.It bootrom , Ns Pa path
 Boot using
 .Ar path
 as a firmware image. Allows booting a UEFI/tianocore binary, such as
 the one from bhyve (source: <https://github.com/freebsd/uefi-edk2.git>; binary <https://people.freebsd.org/~grehan/bhyve_uefi/>).
-.It multiboot , Ns Pa kernel , Ns Oo Pa module Ns Oo ;cmdline Ns Oc Ns Oo :module Ns Oo ;cmdline Ns Oc Ns Oc Ns ... Oc , Ns Oo Pa cmdline Ns Oc
+.It multiboot , Ns Pa kernel Ns Oo , Ns Pa module Ns Oo ;cmdline Ns Oc Ns Oo :module Ns Oo ;cmdline Ns Oc Ns Oc Ns ... Oc Ns Oo , Ns Pa cmdline Ns Oc
```
ijc25: Just below here it says:

> Note that all the commas are required in each case, even if the options are empty.

Which I think is not quite true any more. The commas are still needed for intermediate missing entries, but trailing commas are not. i.e. `kexec,kernel,,"console=foo"` still needs the `,,` for the unused initrd. Whereas `bootrom,foo` no longer needs to be `bootrom,foo,,,`.

Not sure how best to express that, but perhaps simply dropping the words `all the` to leave:

> Note that commas are required in each case, even if the options are empty.

?

> Note that commas are required even for empty options with the exception that they may be omitted for any trailing empty options.

?

src/hyperkit.c
```diff
@@ -745,51 +745,47 @@ firmware_parse(const char *opt) {
...
+	if (opt2 && (cp = strchr(opt2, ',')) != NULL) {
+		*cp = '\0';
+		opt3 = cp + 1;
+	}
 
-		opt2 = strlen(opt2) ? opt2 : NULL;
-		opt3 = strlen(opt3) ? opt3 : NULL;
+	opt1 = opt1 && strlen(opt1) ? opt1 : NULL;
+	opt2 = opt2 && strlen(opt2) ? opt2 : NULL;
+	opt3 = opt3 && strlen(opt3) ? opt3 : NULL;
```
ijc25: I'd replace `strlen(optN)` with `*optN` in each case, we don't need to know the full length, just that it is non-zero.

At first I thought this was a NOP, until I realised it would convert `""` into `NULL`. Perhaps that is worth a comment.

src/hyperkit.c
```diff
@@ -745,51 +745,47 @@ firmware_parse(const char *opt) {
...
-		if ((cp = strchr(fw, ',')) != NULL) {
-			*cp = '\0';
-			opt1 = cp + 1;
-		} else {
-			goto fail;
-		}
+	if ((cp = strchr(fw, ',')) != NULL) {
+		*cp = '\0';
+		opt1 = cp + 1;
+	}
```
ijc25: how about:
```
#define NEXTARG(cur, next, scratch) do {			\
	if cur && scratch = strchr(cur, ',')) != NULL) {	\
		*scratch = '\0';				\
		*next = scratch + 1;				\
    }								\
} while (0)
```
then
```
NEXTARG(fw, opt1);
NEXTARG(opt1, opt2);
NEXTARG(opt2, opt3);
```
?

ijc25:
Couple of comments above (all suggestions really so feel free to say you prefer it how it is).

Ideally we'd support an arbitrary `argc`/`argv` style thing but what you've done here is a worthwhile improvement and there's no obvious/simple POSIX library function to split a string on `,`. So unless you feel super motivated to write a comma separating parser lets just go with this approach for now.

Sadly we can't quite use `wordexp(3)` I think since it splits on spaces not `,` and the nested quoting would get quite "exciting" on the CLI I think.

hyperkit.1
pborzenkov: I like the second one. While it is a bit wordy it does clearly state that the commas may be omitted in some cases.

src/hyperkit.c
```diff
@@ -745,51 +745,44 @@ firmware_parse(const char *opt) {
...
-		} else {
-			goto fail;
-		}
+// Gets next comma-separated option and stores it in next.
+// If the length of the option is 0 it is replaced with NULL.
+#define NEXTARG(cur, next, scratch) do {			\
+	if (cur && (scratch = strchr(cur, ',')) != NULL) {	\
+		*scratch = '\0';				\
+		next = scratch + 1;				\
+		next = next && strlen(next) ? next : NULL;	\
```
ijc25: `next` cannot be `NULL` here (exception some sort of wraparound in `scratch + 1`.

Also wont setting `next = NULL` here cause the next invocation of `NEXTARG` to fail the initial condition and therefore break the `foo,,bar` case because `opt2` ends up NULL and hence `NEXTARGS(opt2, opt3, cp)` fails the `if (cur`.

pborzenkov: Ooops... I haven't thought it through. Shame on me. Will fix.

ijc25:
Thanks!

PR #129, opened 2017-05-10
virtio-vpnkit: remove the strict remote version check

djs55:
Previously we would fail if the remote server didn't report a version of exactly 1. In practice this protocol has evolved only by extension, with new command types being added. All known versions support the "ethernet" command that we need.

This patch removes the strict check of the remote vpnkit protocol version. This allows `hyperkit` to connect to the `com.docker.vmnetd` daemon which comes with Docker for Mac (as well as `vpnkit`)


djs55:
For future reference, this could help with https://github.com/docker/for-mac/issues/155

PR #131, opened 2017-05-16
qcow: update to v0.10.0

djs55:
From hyperkit's point of view, the qcow changes are:

- fix a major performance problem with online GC/compact which could manifest as an I/O timeout, triggering an AHCI reset and then deadlock
- gather performance counters in a format suitable for prometheus (although there is no code to expose them anywhere yet)

https://github.com/mirage/ocaml-qcow/blob/master/CHANGES.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ via `brew` and using `opam` to install the appropriate libraries:
     $ brew install opam libev
     $ opam init
     $ eval `opam config env`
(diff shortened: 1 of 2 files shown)


djs55:
I suspect the upstream `opam-repository` HTTP server might not have been updated -- if so, I'll investigate and rebuild.

ijc25:
LGTM too, once green.

djs55:
I had to add an extra dependency on the `qcow-tool` package since it was split out of `qcow`. CI is now green -- I'll go ahead and merge this now.

PR #132, opened 2017-05-18
qcow: add an optional qcow-stats-config= argument

djs55:
The qcow library gathers a set of metrics in prometheus format including

- read, write, discard rate
- bytes used, file size
- bytes erased, deallocated and in various GC states

It's useful to be able to expose these for monitoring, see for example the Grafana dashboard here:

  https://github.com/mirage/ocaml-qcow/issues/99

This patch adds an optional `qcow-stats-config=` argument which supports values

- `unix:/path/to/socket` -- listen on a Unix domain socket
- `tcp:port` -- listen on `0.0.0.0:port`

(unfortunately the Mirage `conduit` library doesn't allow the bind address to be customised, so care must be taken with the `tcp:` option)

By default nothing is exposed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ FIRMWARE_LIB_SRC := \
 
 HYPERKIT_SRC := src/hyperkit.c
 
-HAVE_OCAML_QCOW := $(shell if ocamlfind query qcow uri logs logs.fmt mirage-unix >/dev/null 2>/dev/null ; then echo YES ; else echo NO; fi)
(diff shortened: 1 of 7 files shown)


ijc25:
C bits seem good to me. Would be nice to have an ocaml person's ack, @samoht perhaps?

djs55:
I force-pushed a small change: I added `prometheus-app` to the package list in `README.md` and `circle.yml`.

djs55:
@samoht good idea, I've force pushed a new version.

samoht:
LGTM

djs55:
Thanks!

PR #133, opened 2017-05-19
go lib: Add support for 9p sockets

MagnusS:
Add support for specifying the path and tag of 9p unix domain sockets
that will be exposed to the guest. This is needed to e.g. expose the
vpnkit /ports directory in the guest.

Also adds a nextSlot counter when building the command line so it's
easier to add new parameters that could require multiple PCI slots.

The sample only supports a single 9p socket for now.

Signed-off-by: Magnus Skjegstad <magnus@skjegstad.com>

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -63,6 +63,12 @@ var defaultHyperKits = []string{"hyperkit",
 	"/Applications/Docker.app/Contents/Resources/bin/hyperkit",
 	"/Applications/Docker.app/Contents/MacOS/com.docker.hyperkit"}
 
+// Socket9P contains a unix domain socket path and 9p tag
+type Socket9P struct {
+	Path string `json:"path"`
+	Tag  string `json:"tag"`
+}
+
 // HyperKit contains the configuration of the hyperkit VM
 type HyperKit struct {
 	// HyperKit is the path to the hyperkit binary
@@ -85,6 +91,9 @@ type HyperKit struct {
 	VSockPorts []int `json:"vsock_ports"`
 	// VSock guest CID
 	VSockGuestCID int `json:"vsock_guest_cid"`
+
+	// 9P sockets
+	Sockets9P []Socket9P `json:"9p_sockets"`
 
 	// Kernel is the path to the kernel image to boot
 	Kernel string `json:"kernel"`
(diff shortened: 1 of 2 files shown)


go/hyperkit.go
```diff
@@ -367,32 +375,49 @@ func (h *HyperKit) buildArgs(cmdline string) {
 	a = append(a, "-m", fmt.Sprintf("%dM", h.Memory))
 
 	a = append(a, "-s", "0:0,hostbridge")
+	a = append(a, "-s", "31,lpc")
+
+	nextSlot := 1
```
rneugeba: I don't think we need this. We can just pick slots and hard code them.

go/hyperkit.go
```diff
@@ -367,32 +375,49 @@ func (h *HyperKit) buildArgs(cmdline string) {
...
 	}
+
 	if h.ISOImage != "" {
-		a = append(a, "-s", fmt.Sprintf("4,ahci-cd,%s", h.ISOImage))
+		a = append(a, "-s", fmt.Sprintf("%d,ahci-cd,%s", nextSlot, h.ISOImage))
+		nextSlot++
 	}
 
-	a = append(a, "-s", "10,virtio-rnd")
-	a = append(a, "-s", "31,lpc")
```
rneugeba: Is there a reason you are dropping `lpc`

go/hyperkit.go
MagnusS: It's to be able to add multiple entries for some parameters, such as disk, or 9p mounts in this case. Without it we would only support one or zero (or a fixed number of dedicated slots).

go/hyperkit.go
MagnusS: It's not dropped - I moved it up to the other fixed slot, in https://github.com/moby/hyperkit/pull/133/commits/0d990789653c070233cfaff3d788c6b2d26268d8#diff-f91b620a5f864931ab2ac3a6cf135a11R378

MagnusS:
Had to fix formatting for CI, will squash the commits after review

go/hyperkit.go
rneugeba: Ah, I missed that

go/hyperkit.go
rneugeba: we don't support multiple disks in the API, not multiple other devices, but I see 9p mounts is an array, but then it's not exposed via the sample CLI. Do you plan to support multiple 9p sockets via `linuxkit run`?

To be clear, I don't have strong objections to making this dynamic.

MagnusS: This change was mostly to make the option I added to the API support the same features as the underlying process and make it easier in the future to support multiple other devices.

My use case is getting Docker for Mac running on Linuxkit and it uses two 9p sockets at the moment. It would be nice to have support in linuxkit for this while testing, but eventually one of the sockets will go away so it's not needed in the longer term.

I don't really have a strong opinion about this either. If you think we don't need it I can remove it from the PR.

rneugeba: I'm just concerned that we add more and more features to the API, but if it will be used for Docker for Mac porting that's fine with me. Should just remember to remove it after because it's like going to bitrod.

rneugeba:
could you squash?

MagnusS:
Yep, squashed and green again

MagnusS:
Thanks!

PR #135, opened 2017-05-31
Improve terminal handling

rneugeba:
For writing LinuxKit tests we would like to redirect the output of the VM to a file (or pipe to `grep`). This was not possible since the `stdio` backend in hyperkit (`-l comZ,stdio`) expects to run on a terminal.

This PR mainly checks if run against a terminal, and if not, uses the PTY backend (`-l comX,autopty=...`). Then we use a go routing to copy the VMs output to stdout. This way the terminal emulation code gets a PTY and we can redirect.

This PR also copies some small pieces of code from containerd to facilitate the Terminal handling and also renames the package to `github.com/moby/hyperkit`


go/hyperkit.go
```diff
@@ -212,6 +213,9 @@ func (h *HyperKit) execute(cmdline string) error {
 	if h.Console == ConsoleFile && h.StateDir == "" {
 		return fmt.Errorf("If ConsoleFile is set, StateDir must be specified")
 	}
+	if h.Console == ConsoleStdio && !isTerminal(os.Stdout) && h.StateDir == "" {
+		return fmt.Errorf("If ConsoleStio is set but stdio is not a terminal, StateDir must be specified")
```
ijc25: Missing a `d` in the first `Stio`.

go/hyperkit.go
```diff
@@ -420,10 +424,10 @@ func (h *HyperKit) buildArgs(cmdline string) {
...
 	}
 
-	if h.Console == ConsoleFile {
-		a = append(a, "-l", fmt.Sprintf("com1,autopty=%s/tty,log=%s/console-ring", h.StateDir, h.StateDir))
-	} else {
+	if h.Console == ConsoleStdio && isTerminal(os.Stdout) {
 		a = append(a, "-l", "com1,stdio")
+	} else if h.StateDir != "" {
+		a = append(a, "-l", fmt.Sprintf("com1,autopty=%s/tty,log=%s/console-ring", h.StateDir, h.StateDir))
 	}
```
ijc25: Is there any way we can get through that if/else chain without doing anything?

go/hyperkit.go
```diff
@@ -439,14 +443,43 @@ func (h *HyperKit) execHyperKit() error {
...
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+		} else {
+			go func() {
+				ttyPath := fmt.Sprintf("%s/tty", h.StateDir)
+				var tty *os.File
+				var err error
+				for {
+					tty, err = os.OpenFile(ttyPath, os.O_RDWR, 0)
+					if err != nil {
```
ijc25: Maybe this should check `os.IsNotExist(err)` and if not fail hard?

go/hyperkit.go
```diff
@@ -439,14 +443,43 @@ func (h *HyperKit) execHyperKit() error {
...
+			cmd.Stdin = os.Stdin
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+		} else {
+			go func() {
+				ttyPath := fmt.Sprintf("%s/tty", h.StateDir)
+				var tty *os.File
+				var err error
+				for {
+					tty, err = os.OpenFile(ttyPath, os.O_RDWR, 0)
```
ijc25: Could be `os.O_RDONLY` I think?

go/pty_util.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func saneTerminal(f *os.File) error {
+	// Go doesn't have a wrapper for any of the termios ioctls.
+	var termios unix.Termios
+	if err := tcget(f.Fd(), &termios); err != nil {
+		return err
+	}
+	// Set -onlcr so we don't have to deal with \r.
+	termios.Oflag &^= unix.ONLCR
```
ijc25: TIL that `&^=` is a thing in Go...

go/hyperkit.go
rneugeba: there may be, but you'd end up with a pretty un-usable system. Without console and without state directory you can't really access the VM.

go/hyperkit.go
rneugeba: I'd have to plumb through a error channel and have `Start()` return an error. None of the function in the wrapper fail hard.

rneugeba:
Fixed some of the comments. What's your preference regarding erroring out if there is a error on opening the tty file?

go/hyperkit.go
ijc25: Should we add an `else { return error.New("oh bother") }` ?

ijc25:
Looks like an error returned at that point would be propagated all the way back out through `Run` or `Start` as needed already (maybe I'm confused though since you seemed to think not), so I think I'd prefer to return errors other than `ENOEXIST` (which should loop, ideally there would be a timeout, but no biggie).

rneugeba:
I may also be confused...the code is executed in a go routine and I frankly don't know how a return would be handled/propagated (need to read up on Go a bit more)

ijc25:
> the code is executed in a go routine

Ah that does complicate things somewhat! You'd need to do something with cancellable contexts and/or `chan error` type things I think. Almost certainly not worth the effort to rearchitect at this point. Although I should note that if a long running daemon were to use this code it could potentially leak goroutines until it exhausted some resource or other, but that's an even bigger `context`ification job.

ijc25:
I have no other comments so LGTM now.

rneugeba:
thanks for the review. I'll merge this now. I think some of the conditionals in the check can and probably should be streamlined. I may do a separate PR for that

PR #138, opened 2017-06-07
go: Make hyperkit wrapper compilable on other OSes

rneugeba:
In LinuxKit we want to use the HyperKit wrapper on all supported
OSes and just error out when not running on macOS.

The previous commit (terminal handling) introduced macOS specifics
which fail to compile for Linux. This commit adds a fallback
for OSes other than macOS, which simply error out when called.


justincormack:
Oh I dont seem to have write access any more...

justincormack:
Can someone give me back admin access as I am a maintainer...

rneugeba:
@justincormack gave you back admin right (i had to request it too after the move)

ijc25:
@rneugeba how hard would it be to either add a Linux (and others?) cross compile test to the existing `circle.yml` or to add a Linux CI (travis or something)?

rneugeba:
@ijc25 easy: https://github.com/moby/hyperkit/pull/139

PR #139, opened 2017-06-08
go: Add cross-compile tests for Linux and Windows

rneugeba:
While the hyperkit go package is only useful on macOS, consumers
may want to use it in code which also also needs compiling on Windows
and Linux without having to use OS conditional compilation.

This patch adds checks to the CI to ensure the code cross compiles
on other platforms.


ijc25:
LGTM. Merging.

PR #141, opened 2017-06-13
go API: Add support for more than one block device

### LINKED ISSUE #137: support multiple disks in Go API
We want to be able to support attaching multiple disks from the Go API for LinuxKit, currently we only do for qemu. 

This is needed to fix https://github.com/moby/tool/issues/68 and make all the tests more portable.

Plus it is just a useful feature.

MagnusS:
Changes DiskImage to an array called DiskImages with keys Path and
SizeMB. Also changes the JSON key from "disk" to "disk_images".
DiskSize is removed as size is set per disk image.

This allows multiple block devices to be mounted via the API and fixes #137

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -47,7 +47,6 @@ const (
 	ConsoleFile
 
 	defaultVPNKitSock = "Library/Containers/com.docker.docker/Data/s50"
-	defaultDiskImage  = "disk.img"
 
 	defaultCPUs   = 1
 	defaultMemory = 1024 // 1G
@@ -70,6 +69,12 @@ type Socket9P struct {
 	Tag  string `json:"tag"`
 }
 
+// DiskConfig contains the path to a disk image and an optional size if the image needs to be created.
+type DiskConfig struct {
+	Path string `json:"path"`
+	Size int    `json:"size"`
+}
+
 // HyperKit contains the configuration of the hyperkit VM
 type HyperKit struct {
 	// HyperKit is the path to the hyperkit binary
@@ -82,8 +87,8 @@ type HyperKit struct {
 	VPNKitKey string `json:"vpnkit_key"`
 	// UUID is a string containing a UUID, it sets BIOS DMI UUID for the VM (as found in /sys/class/dmi/id/product_uuid on Linux).
 	UUID string `json:"uuid"`
-	// DiskImage is the path to the disk image to use
-	DiskImage string `json:"disk"`
+	// Disks contains disk images to use/create.
+	Disks []DiskConfig `json:"disks"`
 	// ISOImage is the (optional) path to a ISO image to attach
 	ISOImage string `json:"iso"`
 	// VSock enables the virtio-socket device and exposes it on the host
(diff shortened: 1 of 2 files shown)


go/hyperkit.go
```diff
@@ -70,6 +69,12 @@ type Socket9P struct {
 	Tag  string `json:"tag"`
 }
 
+// DiskImage contains the path to a disk image and an optional size if the image needs to be created.
+type DiskImage struct {
+	Path   string `json:"path"`
+	SizeMB int    `json:"size"`
```
rn: Can we call this `DiskConfig` and `Size` in the same way as https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/util.go#L139
The code is in different repos, but might as well try to keep it consistent

go/hyperkit.go
```diff
@@ -82,8 +87,8 @@ type HyperKit struct {
 	VPNKitKey string `json:"vpnkit_key"`
 	// UUID is a string containing a UUID, it sets BIOS DMI UUID for the VM (as found in /sys/class/dmi/id/product_uuid on Linux).
 	UUID string `json:"uuid"`
-	// DiskImage is the path to the disk image to use
-	DiskImage string `json:"disk"`
+	// DiskImages contains disk images to use/create.
+	DiskImages []DiskImage `json:"disk_images"`
```
rn: Maybe make this: `Disks []DiskConfig` like here: https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_qemu.go#L27

go/sample/main.go
```diff
@@ -33,7 +33,7 @@ func main() {
 	hk := flag.String("hyperkit", "", "HyperKit binary to use")
 	statedir := flag.String("state", "", "Directory to keep state in")
 	vpnkitsock := flag.String("vpnkitsock", "auto", "Path to VPNKit socket")
-	disk := flag.String("disk", "", "Path to disk image")
+	disks := flag.String("disk", "", "Comma separated list of disk images, format [image]{:size_in_MB}. If size is specified the image will be created if it doesn't exist. If image name is omitted a default name will be used.")
```
rn: Could we maybe make this the same as/similar to the LinuxKit command line, as in https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_qemu.go#L64

this also has a more elgant way for parsing the command line in: https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/util.go#L153 so the code below can be moved into a method

MagnusS: Yeah I just added something simple in the sample. But it's nice to be consistent - I'll update

MagnusS:
@rn thanks for the review. I've updated to make it more consistent with current linuxkit.

go/sample/main.go
```diff
@@ -101,9 +100,14 @@ func main() {
 		}
 		h.VSockPorts = ports
 	}
-	h.DiskImage = *disk
 	h.ISOImage = *iso
 
+	if len(disks) == 0 {
+		log.Fatalln("No disk specified")
```
justincormack: Its fine to specify no disks surely?

justincormack:
Minor comment but LGTM

go/sample/main.go
MagnusS: Yes - updated

PR #143, opened 2017-06-19
virtio: Fix device ID for the virtio random device

rn:
According to the virtio spec[1] the device ID for the virtio
entropy source is 0x1005 and not 0x1002. This fix was already applied
to bhyve[2]. Linux and FreeBSD guests typically match on the
device type and ignore the device ID.

[1] http://docs.oasis-open.org/virtio/virtio/v1.0/cs04/virtio-v1.0-cs04.html#x1-6600011
[2] https://github.com/freebsd/freebsd/commit/919dec09c25924259b40d9f6d09a52ac00bbbba2

### DIFF
--- a/src/include/xhyve/virtio.h
+++ b/src/include/xhyve/virtio.h
@@ -212,7 +212,7 @@ struct vring_used {
 #define	VIRTIO_VENDOR		0x1AF4
 #define	VIRTIO_DEV_NET		0x1000
 #define	VIRTIO_DEV_BLOCK	0x1001
-#define	VIRTIO_DEV_RANDOM	0x1002
+#define	VIRTIO_DEV_RANDOM	0x1005
 #define VIRTIO_DEV_9P		0x1009
 #define VIRTIO_DEV_SOCK		0x103f /* In the legacy range. */
 


ijc:
IIRC for "legacy" (pre v1.0) virtio implementations (which is what hyperkit implements) the device ID is arbitrary and some other field (I forget which) is used to make the binding between driver and device, but it is valid, reasonable and a little less confusing to use the post-standardization numbers regardless.

My only minor concern is that I'm sure I wrote this patch at some point and then dropped it for some reason, but I can't remember at all why or if it was important. Lets assume that given I don't remember it can't have been all that important.

So, LGTM.

ijc:
CI failure is 
```
ld: warning: object file (/usr/local/lib/ocaml/stdlib.a(weak.o)) was built for newer OSX version (10.11) than being linked (10.10)
ld hyperkit.sym
clang: error: no such file or directory: '/Users/distiller/.opam/system/lib/io-page/io_page_unix.a'
clang: error: no such file or directory: '/Users/distiller/.opam/system/lib/io-page/libio_page_unix_stubs.a'
```
There are lots of the 10.10 vs 10.11 messages.

I thought @djs55 had sorted the version thing ages ago, and IIRC was also looking at the ocaml linking stuff recently.

ijc:
@rn I think if you rebase onto #144 your CI issues will go away.

rn:
@ijc @djs55 thanks for fixing. I've updated and now the compile passes. merging

PR #144, opened 2017-06-20
ocaml: the io_page_unix stubs are in a different directory

djs55:
As a result of the recent `io-page` refactor which split the package into 3 to remove the optional dependency and simplify the build rules more generally[1], the stubs are now in the directory `io-page-unix`.

This should fix the CircleCI build.

[1] https://discuss.ocaml.org/t/ann-io-page-2-0-0-with-packaging-changes/440


ijc:
LGTM once CI agrees.

PR #150, opened 2017-08-02
Add support for setting a bootrom in Go API

justincormack:
This can be used for UEFI boot for example.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -110,6 +110,8 @@ type HyperKit struct {
 	Kernel string `json:"kernel"`
 	// Initrd is the path to the initial ramdisk to boot off
 	Initrd string `json:"initrd"`
+	// Bootrom is the path to a boot rom eg for UEFI boot
+	Bootrom string `json:"bootrom"`
 
 	// CPUs is the number CPUs to configure
 	CPUs int `json:"cpus"`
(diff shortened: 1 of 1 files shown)


justincormack:
I have a LinuxKit patch to support UEFI ISO boot on hyperkit once this is merged.

justincormack:
Argh, Ocaml has broken
```
type_conv.109.15.00 is not available because your system doesn't comply with ocaml-version >= "4.00.1" & ocaml-version < "4.03.0".
```

cc @djs55

samoht:
Seems that brew updated the version of OCaml to 4.05:

```
Downloading https://homebrew.bintray.com/bottles/ocaml-4.05.0.el_capitan.bot
```

A possible fix would be to replace https://github.com/moby/hyperkit/blob/master/circle.yml#L21 by
```
opam init --yes --compiler=4.04.2
```

djs55:
@samoht thanks -- I'll give that a go. Looking at the error message I might also try to see if I can pin a hash of `cohttp` -- maybe the problem would be fixed in the unreleased 1.0?

djs55:
I've updated some dependencies build with OCaml 4.05. Once these are merged:

- https://github.com/ocaml/opam-repository/pull/10001
- https://github.com/ocaml/opam-repository/pull/10002

I'll make a PR to fix the CI here. Hopefully it'll work with old and new versions of OCaml.

djs55:
@samoht I've got another one for you! https://github.com/ocaml/opam-repository/pull/10007

samoht:
@djs55 I've merged it as well

djs55:
Fix for the CI is in #151

PR #151, opened 2017-08-03
ocaml: update packages to fix CI with OCaml 4.05

djs55:
`brew` now has OCaml 4.05 which is incompatible with some of the older dependencies. This patch bumps the versions of some of the dependencies and updates the linking step to match.

As a special bonus this removes some libraries which were being linked by accident and shrinks the binary size from:

```
$ ls -l build/hyperkit
-rwxr-xr-x  1 djs  staff  7356796  3 Aug 17:25 build/hyperkit
```
to
```
$ ls -l build/hyperkit
-rwxr-xr-x  1 djs  staff  3965068  3 Aug 17:00 build/hyperkit 
```

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,7 @@ test:
     - make test
     - brew install opam libev
     - opam init --yes
-    - opam install --yes uri qcow.0.10.0 qcow-tool mirage-block-unix.2.7.0 ocamlfind conf-libev logs fmt mirage-unix prometheus-app
+    - opam install --yes uri qcow.0.10.3 qcow-tool mirage-block-unix.2.7.0 ocamlfind conf-libev logs fmt mirage-unix prometheus-app
     - make clean
     - eval `opam config env` && make all
     - make test
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,7 @@ test:
     - make test
     - brew install opam libev
     - opam init --yes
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -11,19 +11,21 @@ machine:
...
-    - brew install opam libev
-    - opam init --yes
-    - opam install --yes uri qcow.0.10.0 qcow-tool mirage-block-unix.2.7.0 ocamlfind conf-libev logs fmt mirage-unix prometheus-app
-    - make clean
-    - eval `opam config env` && make all
-    - make test
-    - eval `opam config env` && make test-qcow
+    - brew install opam
+    - opam init
+    - opam config exec -- opam depext -i uri qcow.0.10.3 conduit.1.0.0 lwt.3.1.0 qcow-tool mirage-block-unix.2.7.0 ocamlfind conf-libev logs fmt mirage-unix prometheus-app
```
samoht: normally you don't need this, `opam depext` will run in the correct context by default

justincormack:
Merginf to unblock CI, can tweak later if wanted...

circle.yml
djs55: That's what I thought too. I was having strange `PATH` problems (I think) where the `opam-depext` command couldn't be found. After this is merged I'll make a PR to remove this bit and see what happens.

PR #154, opened 2017-08-30
VPNKit protocol and Go API update

MagnusS:
This updates the VPNKit protocol to version 22. This change is incompatible with earlier versions of VPNKit and depends on moby/vpnkit#277.

It was previously possible to encode an IP address in the UUID to request a fixed address from VPNKit. This replaces this functionality with a `preferred_ipv4=` option that can be passed to the driver. IPs encoded in UUIDs are now ignored and treated as a regular UUID.

The new server error messages allows the driver to inform the user about some common VPNKit error conditions, such as if the IP is unavailable. The old driver would just fail with an unspecified vif error.

### DIFF
--- a/src/lib/pci_virtio_net_vpnkit.c
+++ b/src/lib/pci_virtio_net_vpnkit.c
@@ -77,6 +77,8 @@
 #include <xhyve/pci_emul.h>
 #include <xhyve/mevent.h>
 #include <xhyve/virtio.h>
+#include <arpa/inet.h>
+#include <libkern/OSByteOrder.h>
 
 #define WPRINTF(format, ...) printf(format, __VA_ARGS__)
 
@@ -93,13 +95,18 @@ struct msg_init {
 } __packed;
 
 #define CMD_ETHERNET 1
-struct msg_common {
-	uint8_t command;
+#define CMD_PREFERRED_IPV4 8
+
+#define RESP_VIF 1
+#define RESP_DISCONNECT 2
+
+struct cmd_ethernet {
+	char uuid[36];
 } __packed;
 
-struct msg_ethernet {
-	uint8_t command; /* CMD_ETHERNET */
+struct cmd_preferred_ipv4 {
 	char uuid[36];
+	in_addr_t ip;
 } __packed;
 
 struct vif_info {
@@ -108,6 +115,26 @@ struct vif_info {
 	uint8_t mac[6];
 } __packed;
 
+struct disconnect_reason {
+	uint8_t len;
+	char msg[256];
+} __packed;
+
+struct msg_command {
+	uint8_t command;
+	union {
+		struct cmd_ethernet ethernet;
+		struct cmd_preferred_ipv4 preferred_ipv4;
+	};
+} __packed;
+
+struct msg_response {
+	uint8_t response_type;
+	union {
+		struct disconnect_reason disconnect;
+		struct vif_info vif;
+	};
+} __packed;
 
 /*
  * Host capabilities.  Note that we only offer a few of these.
@@ -286,11 +313,11 @@ err:
 /*
  * wire protocol
  */
-static int vpnkit_connect(int fd, const char uuid[36], struct vif_info *vif)
+static int vpnkit_connect(int fd, const char uuid[36], struct vif_info *vif, in_addr_t preferred_ipv4)
 {
 	struct msg_init init_msg = {
 		.magic = { 'V', 'M', 'N', '3', 'T' },
-		.version = 1U,
+		.version = 22U,
 	};
 
 	/* msg.commit is not NULL terminated */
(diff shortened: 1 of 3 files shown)


rn:
Is the new VPNKit already shipping with Docker for Mac? Should we only merge this once it is at least in `-edge`?

MagnusS:
@rn not yet, as we need an updated `hyperkit` binary as well. I can use the binary from this PR if necessary though. If we merge you wouldn't be able to run `hyperkit` master with `vpnkit` without updating that from master as well, but tagged releases of `vpnkit`/`hyperkit` could still be used with stable/edge. We should probably wait with updating the go api in `linuxkit` though, as the `--ip` option would stop working until there's a new `hyperkit` in D4M.

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -324,21 +343,60 @@ static int vpnkit_connect(int fd, const char uuid[36], struct vif_info *vif)
...
+
+		if (really_write(fd, (uint8_t*)&cmd_ethernet, sizeof(cmd_ethernet)) < 0) {
+			fprintf(stderr, "virtio-net-vpnkit: failed to write ethernet cmd\n");
+			return -1;
+		}
+
+	} else { // preferred_ipv4
+		struct msg_preferred_ipv4 cmd_preferred_ipv4 = {
+			.command = CMD_PREFERRED_IPV4,
+			.ip = OSSwapInt32(*preferred_ipv4), // VPNKit uses LE, our IP is in network byte order
```
rn: Is VPNKit explicitly in LE or is it in host byte order. If the latter this should be `htonl` or equiv

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -324,21 +343,60 @@ static int vpnkit_connect(int fd, const char uuid[36], struct vif_info *vif)
...
-		return -1;
+	if (init_reply.version != init_msg.version) { 
+		fprintf(stderr, "virtio-net-vpnkit: protocol version mismatch: version %d requested, got version %d.\n", 
+			init_msg.version,
+			init_reply.version);
 	}
 
-	if (really_read(fd, (uint8_t*)vif, sizeof(*vif)) < 0) {
-		fprintf(stderr, "virtio-net-vpnkit: failed to read vif info\n");
+	if (preferred_ipv4 == NULL) { // ethernet
```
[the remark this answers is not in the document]
ijc: TBH I'm not really sure what this comment is supposed to tell me, regardless of where it is placed.

I'd prefer if the if/else was inverted so it was `preferred_ipv4 != NULL { do stuff using preferred_ipv4 } else { do the other thing }`. With that I also think there would be no need for a comment for the `preferred_ipv4 != NULL` case since it would be self explanatory, the `else` case would have a `/* If no preferred ipv4 address configured then we «explain the fallback» */`

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -324,21 +343,60 @@ static int vpnkit_connect(int fd, const char uuid[36], struct vif_info *vif)
...
 
+	struct disconnect_reason *reason;
+
+	switch (reply.response_type) {
+		case RESP_DISCONNECT: 
+			reason = (struct disconnect_reason*)&(reply.data);
+			char buf[257];
+			memcpy(&buf[0], &(reason->msg[0]), reason->len);
+			buf[reason->len] = '\0';
+			fprintf(stderr, "virtio-net-vpnkit: server disconnected: %s\n", buf);
```
ijc: I wonder if this works:
```
fprintf(stderr, "virtio-net-vpnkit: server disconnected: %*s\n", reason->len, (casting stuff)reason->msg)
```
?

If that doesn't work then I'm a little bit unhappy with the hardcoded assumption about the range of `reason->len` being encoded in that `char buf[257]` and hence the overflow check being implicit and (more importantly) based on code far away. An `assert(sizeof(reason->len) == 1)` would be one approach, or something like `char buf[sizeof(reason->msg)+1]` might work?

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -423,7 +493,7 @@ vpnkit_create(struct pci_vtnet_softc *sc, const char *opts)
 			goto err;
 		}
 
-		if (vpnkit_connect(fd, uuid_string, &state->vif) == 0)
+		if (vpnkit_connect(fd, uuid_string, &state->vif, (ipv4 == NULL ? NULL : &preferred_ipv4.s_addr)) == 0)
```
ijc: Perhaps you could arrange for `s_addr` to be zero being the default/unconfigured action, which would avoid the conditionals and pointer argument here, i.e. vpnkit_connect could just take a plain `in_addr_t`.

src/lib/pci_virtio_net_vpnkit.c
MagnusS: VPNKit explicitly uses LE

src/lib/pci_virtio_net_vpnkit.c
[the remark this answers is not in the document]
MagnusS: Hm, yes - I agree that would be more readable, I'll swap them around and remove the pointer as suggested below

src/lib/pci_virtio_net_vpnkit.c
MagnusS: That worked! I'll remove the buffer

MagnusS:
@rn @ijc thanks for the reviews! I've pushed an update.

Re timing, I think we could merge this when we are ready to include it in D4D master/nightly (tests are green, 17.07 is released etc), which is hopefully by end of this week. We would then have to wait a few weeks with updating the go vendoring in linuxkit though (or update and break`--ip` temporarily, unless you run the nightly build). Do you think what would work?

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -93,13 +95,21 @@ struct msg_init {
...
+#define RESP_VIF 1
+#define RESP_DISCONNECT 2
 
 struct msg_ethernet {
 	uint8_t command; /* CMD_ETHERNET */
 	char uuid[36];
+	uint32_t unused;
+} __packed;
+
+struct msg_preferred_ipv4 {
```
rn: Are the messages always the same size? Looks like `msg_ethernet` and `msg_preferred_ipv4` are. If so, a `union` will be a better structure for this.

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -324,19 +343,56 @@ static int vpnkit_connect(int fd, const char uuid[36], struct vif_info *vif)
...
-		fprintf(stderr, "virtio-net-vpnkit: failed to read vif info\n");
-		return -1;
+	struct disconnect_reason *reason;
+
+	switch (reply.response_type) {
+		case RESP_DISCONNECT: 
+			reason = (struct disconnect_reason*)&(reply.data);
+			fprintf(stderr, "virtio-net-vpnkit: server disconnected: %*s\n", reason->len, reason->msg);
+			return -1;
+		case RESP_VIF:
```
rn: I would move this to the top as this is a the normal/common case which indicates success. `RESP_DISCONNECT` and "unknown" are both error cases and should be grouped together

src/lib/pci_virtio_net_vpnkit.c
```diff
@@ -108,6 +118,15 @@ struct vif_info {
 	uint8_t mac[6];
 } __packed;
 
+struct disconnect_reason {
+	uint8_t len;
+	char msg[256];
+} __packed;
+
+struct msg_response {
```
rn: I would move this declaration above `disconnect_reason` as this is what we always get. and then `disconnect_reason` is would be in data.

Also, it might make sense to make this a union where it's either `data` or `disconnect_reason` with `response_type` the selector.

MagnusS:
@rn thanks - the struct's are the same size, so I updated to use unions and moved the case. I left the extra commit for review, but squashed and force pushed now.

MagnusS:
Merging this now as all tests passed on the build from this PR in D4D. A D4D build with updated vpnkit/hyperkit should be available soon.

MagnusS:
This feature is going into 17.09, so it will be in the edge rc1 release today/tomorrow. I'll make a PR to update linuxkit

rn:
I don't know how many linuxkit/hyperkit users we have, but possibly a few. What's the error when running linuxkit with these updates against an incompatible DfM? is it sensible? Is there a simple way to fall back to the old protocol (i had a quick look and it didn't seem to be)?

MagnusS:
If you run an updated linuxkit + an older DfM you will get an `invalid option` error if you use `--ip`, but it should otherwise work. 

If you run an outdated linuxkit with an updated DfM release, the `--ip` parameter will be silently ignored as the ip is not parsed from the UUID anymore.

PR #157, opened 2017-10-07
go: create disks with Truncate

djs55:
Previously we made a disk be the correct size by explicitly writing zeroes to it.

This patch uses `Truncate` to make the disk the correct size. If the filesystem supports sparse files (e.g. APFS) then

- it's very fast
- it doesn't actually allocate any space until the VM writes to it


djs55:
Yes -- I double-checked by mounting an HFS+ volume. It was slow but it did work as expected.

PR #158, opened 2017-10-08
Support TRIM on raw files via F_PUNCHHOLE

djs55:
The macOS API `fcntl(F_PUNCHHOLE)` will successfully punch holes in files on APFS and free the backing storage so we can use this to implement the host side of TRIM.

The arguments to the `fcntl` need to be aligned to the filesystems `f_bsize` as returned by `statfs` (which on my Mac's SSD is 4096).

Unfortunately exposing the 4096 byte sectors to the guest seems to trigger a bug elsewhere in the ACHI implementation so this patch maintains a virtual sector size of 512 bytes. This means that the guest will send 512 byte-aligned requests, we must round these to (eg) 4096 byte-aligned requests and manually zero any leading and trailing sectors. For the zeroing we allocate a `delete_zero_buf` per block device and pre-zero it.

When the device is opened we perform a test `F_PUNCHHOLE` of length 0 at offset 0 and check the return value. If the call succeeds then we can assume sparse files is supported (e.g. APFS) and if the call fails we assume sparse files are not supported (e.g. HFS+)

This PR includes 2 patches from bhyve which we were missing. I don't think these are necessary but I thought I'd include them since they exist.

### DIFF
--- a/src/lib/block_if.c
+++ b/src/lib/block_if.c
@@ -33,6 +33,7 @@
 #include <sys/stat.h>
 #include <sys/ioctl.h>
 #include <sys/disk.h>
+#include <sys/mount.h>
 
 #include <assert.h>
 #include <fcntl.h>
@@ -52,6 +53,14 @@
 #include "mirage_block_c.h"
 
 #define BLOCKIF_SIG	0xb109b109
+
+#ifndef ALIGNUP
+#define ALIGNUP(x, a) (((x - 1) & ~(a - 1)) + a)
+#endif
+
+#ifndef ALIGNDOWN
+#define ALIGNDOWN(x, a) (-(a) & (x))
+#endif
 
 /* xhyve: FIXME
  *
@@ -101,6 +110,8 @@ struct blockif_ctxt {
 	off_t			bc_size;
 	int			bc_sectsz;
 	int			bc_psectsz;
+	size_t			bc_delete_alignment; /* f_bsize always a power of 2 */
+	void*       		bc_delete_zero_buf;
 	int			bc_psectoff;
 	int			bc_closing;
 	pthread_t		bc_btid[BLOCKIF_NUMTHR];
@@ -211,6 +222,7 @@ static int
 block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
 {
 	int ret = -1;
+
 #ifdef __FreeBSD__
 	off_t arg[2] = { offset, len };
 #endif
@@ -221,22 +233,52 @@ block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
 		errno = EOPNOTSUPP;
 	else if (bc->bc_rdonly)
 		errno = EROFS;
-	if (bc->bc_fd >= 0) {
-		if (bc->bc_ischr) {
-#ifdef __FreeBSD__
-			ret = ioctl(bc->bc_fd, DIOCGDELETE, arg);
-#else
-			errno = EOPNOTSUPP;
-#endif
-		} else
-			errno = EOPNOTSUPP;
 #ifdef HAVE_OCAML_QCOW
-	} else if (bc->bc_mbh >= 0) {
+	if (bc->bc_mbh >= 0) {
 		ret = mirage_block_delete(bc->bc_mbh, offset, len);
+		goto out;
+	}
 #endif
-	} else
-		abort();
+#ifdef __FreeBSD__
+	if ((bc->bc_fd >= 0) && (bc->bc_ischr)) {
+		ret = ioctl(bc->bc_fd, DIOCGDELETE, arg);
+		errno = EOPNOTSUPP;
+		goto out;
+	}
+#elif __APPLE__
+	if (bc->bc_fd >= 0) {
+		/* PUNCHHOLE lengths and offsets have to be aligned so explicitly write zeroes
+		   into the unaligned parts at the beginning and the end. This wouldn't be necessary
+		   if the host and guest had the same sector size */
+		assert (offset >= 0);
+		assert (len >= 0);
+		size_t fp_offset = (size_t) offset;
+		size_t fp_length = (size_t) len;
 
+		size_t aligned_offset = ALIGNUP(fp_offset, bc->bc_delete_alignment);
+		if (aligned_offset != fp_offset) {
+			size_t len_to_zero = aligned_offset - fp_offset;
+			assert(len_to_zero <= bc->bc_delete_alignment);
+			ssize_t written = pwrite(bc->bc_fd, bc->bc_delete_zero_buf, len_to_zero, (off_t) fp_offset);
+			if (written == -1) goto out;
+			fp_offset += len_to_zero;
+			fp_length -= len_to_zero;
+		}
+		size_t aligned_length = ALIGNDOWN(fp_length, bc->bc_delete_alignment);
+		if (aligned_length != fp_length) {
+			size_t len_to_zero = fp_length - aligned_length;
+			assert(len_to_zero <= bc->bc_delete_alignment);
+			fp_length -= len_to_zero;
+			ssize_t written = pwrite(bc->bc_fd, bc->bc_delete_zero_buf, len_to_zero, (off_t) (fp_offset + fp_length));
+			if (written == -1) goto out;
+		}
+		struct fpunchhole arg = { .fp_flags = 0, .reserved = 0, .fp_offset = (off_t) fp_offset, .fp_length = (off_t) fp_length };
+		ret = fcntl(bc->bc_fd, F_PUNCHHOLE, &arg);
+		goto out;
+	}
+#endif
+	errno = EOPNOTSUPP;
+out:
 	HYPERKIT_BLOCK_DELETE_DONE(offset, ret);
 	return ret;
 }
@@ -554,9 +596,12 @@ blockif_open(const char *optstr, const char *ident)
 	char *nopt, *xopts, *cp;
 	struct blockif_ctxt *bc;
 	struct stat sbuf;
+	struct statfs fsbuf;
 	// struct diocgattr_arg arg;
 	off_t size, psectsz, psectoff, blocks;
 	int extra, fd, i, sectsz;
+	size_t delete_alignment;
+	void *delete_zero_buf;
 	int nocache, sync, ro, candelete, geom, ssopt, pssopt;
 	mirage_block_handle mbh;
 	int use_mirage = 0;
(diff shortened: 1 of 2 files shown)


src/lib/pci_ahci.c
```diff
@@ -977,6 +985,7 @@ handle_identify(struct ahci_port *p, int slot, uint8_t *cfis)
 		buf[88] = 0x7f;
 		if (p->xfermode & ATA_UDMA0)
 			buf[88] |= (1 << ((p->xfermode & 7) + 8));
+		buf[93] = (1 | 1 <<14);
```
ijc: The commit `Support TRIM on raw files via F_PUNCHHOLE` which does this is lacking a `cherry-picked from` note, could you add it please?

src/lib/block_if.c
```diff
@@ -221,22 +228,46 @@ block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
...
+		ret = ioctl(bc->bc_fd, DIOCGDELETE, arg);
+		errno = EOPNOTSUPP;
+		goto out;
+	}
+#elif __APPLE__
+	if (bc->bc_fd >= 0) {
+		/* PUNCHHOLE lengths and offsets have to be aligned so explicitly write zeroes
+		   into the unaligned parts at the beginning and the end. This wouldn't be necessary
+		   if the host and guest had the same sector size */
+		offset_unaligned_by = arg.fp_offset - (arg.fp_offset / bc->bc_delete_alignment * bc->bc_delete_alignment);
```
ijc: Could you encapsulate `arg.fp_offset / bc->bc_delete_alignment * bc->bc_delete_alignment` into a `ROUND_DOWN(arg.fp_offset, bc->bc_delete_alignment)` please. You might be able to steal `src/lib/firmware/kexec.c:#define ALIGNDOWN(x, a) (-(a) & (x))`? (I assume `bc_delete_alignment` is a power of two?)

ijc:
```
src/lib/block_if.c:220:20: error: variable has incomplete type 'struct fpunchhole'
        struct fpunchhole arg = { .fp_flags = 0, .reserved = 0, .fp_offset = offset, .fp_length = len };
                          ^
src/lib/block_if.c:220:9: note: forward declaration of 'struct fpunchhole'
        struct fpunchhole arg = { .fp_flags = 0, .reserved = 0, .fp_offset = offset, .fp_length = len };
               ^
src/lib/block_if.c:265:26: error: use of undeclared identifier 'F_PUNCHHOLE'
                ret = fcntl(bc->bc_fd, F_PUNCHHOLE, &arg);
                                       ^
src/lib/block_if.c:701:22: error: variable has incomplete type 'struct fpunchhole'
                        struct fpunchhole arg = { .fp_flags = 0, .reserved = 0, .fp_offset = 0, .fp_length = 0 };
                                          ^
src/lib/block_if.c:701:11: note: forward declaration of 'struct fpunchhole'
                        struct fpunchhole arg = { .fp_flags = 0, .reserved = 0, .fp_offset = 0, .fp_length = 0 };
                               ^
src/lib/block_if.c:702:18: error: use of undeclared identifier 'F_PUNCHHOLE'
                        if (fcntl(fd, F_PUNCHHOLE, &arg) == 0) {
                                      ^
4 errors generated.
```

Might we need a High Sierra toolchain in CI?

justincormack:
punchhole is in the headers on Sierra (which has beta support for APFS) so maybe just upgrade the that as a build chain now, rather than going full high sierra?

justincormack:
We should fix 4k sectors but transition may well involve reformatting the disk...

ijc:
Perhaps bumping this fragment of `circle.yml` will be sufficient?
```
machine:
  xcode:
    version: "7.3"
```

Hopefully our use of ` MACOSX_DEPLOYMENT_TARGET: "10.10"` is no impediment, bumping it to `10.11` would be OK but probably not further at this point.

justincormack:
10.10 support was only best effort, its probably ok to bump to 10.11

src/lib/block_if.c
```diff
@@ -101,6 +110,8 @@ struct blockif_ctxt {
 	off_t			bc_size;
 	int			bc_sectsz;
 	int			bc_psectsz;
+	int			bc_delete_alignment; /* f_bsize always a power of 2 */
```
ijc: If you make this a `size_t` then I think a bunch of the explicit casting can go away.

src/lib/block_if.c
```diff
@@ -213,6 +224,10 @@ block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
 	int ret = -1;
 #ifdef __FreeBSD__
 	off_t arg[2] = { offset, len };
+#elif __APPLE__
+	struct fpunchhole arg = { .fp_flags = 0, .reserved = 0, .fp_offset = offset, .fp_length = len };
+	size_t len_to_zero;
+	ssize_t aligned_offset, aligned_length, written;
```
ijc: Of these three I think the two `aligned_*` can be `size_t`. If there is any chance of underflow then that ought to be checked before the calculation, and I think because both come from `ALIGN*` there is no such possibility.

src/lib/block_if.c
```diff
@@ -221,22 +236,46 @@ block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
...
+	}
+#elif __APPLE__
+	if (bc->bc_fd >= 0) {
+		/* PUNCHHOLE lengths and offsets have to be aligned so explicitly write zeroes
+		   into the unaligned parts at the beginning and the end. This wouldn't be necessary
+		   if the host and guest had the same sector size */
+		aligned_offset = ALIGNUP(arg.fp_offset, bc->bc_delete_alignment);
+		if (aligned_offset != arg.fp_offset) {
+			len_to_zero = (size_t) (aligned_offset - arg.fp_offset);
+			assert(len_to_zero <= (size_t) bc->bc_delete_alignment);
```
ijc: If you change the type then both these casts can go, I think.

src/lib/block_if.c
```diff
@@ -221,22 +236,46 @@ block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
...
+			written = pwrite(bc->bc_fd, bc->bc_delete_zero_buf, len_to_zero, arg.fp_offset);
+			if (written == -1) goto out;
+			arg.fp_offset += len_to_zero;
+			arg.fp_length -= len_to_zero;
+		}
+		aligned_length = ALIGNDOWN(arg.fp_length, bc->bc_delete_alignment);
+		if (aligned_length != arg.fp_length) {
+			len_to_zero = (size_t) (arg.fp_length - aligned_length);
+			assert(len_to_zero <= (size_t) bc->bc_delete_alignment);
```
ijc: If you change the type both these cases can go.

src/lib/block_if.c
```diff
@@ -554,9 +593,11 @@ blockif_open(const char *optstr, const char *ident)
 	char *nopt, *xopts, *cp;
 	struct blockif_ctxt *bc;
 	struct stat sbuf;
+	struct statfs fsbuf;
 	// struct diocgattr_arg arg;
 	off_t size, psectsz, psectoff, blocks;
-	int extra, fd, i, sectsz;
+	int extra, fd, i, sectsz, delete_alignment;
```
ijc: Make delete_alignment a `size_t` here too.

src/lib/block_if.c
```diff
@@ -655,7 +698,33 @@ blockif_open(const char *optstr, const char *ident)
 			perror("Could not stat backing file");
 			goto err;
 		}
+		if (fstatfs(fd, &fsbuf) < 0) {
+			perror("Could not stat backfile file filesystem");
+			goto err;
+		}
+		delete_alignment = (int)fsbuf.f_bsize;
```
ijc: I think `f_bsize` is a `long`, so you may need to cast to `size_t` here.

src/lib/block_if.c
```diff
@@ -655,7 +698,33 @@ blockif_open(const char *optstr, const char *ident)
...
+				perror("fcntl(F_PUNCHHOLE) failed: host filesystem does not support sparse files");
+				candelete = 0;
+			}
+		}
+#endif
 	}
+	/* delete_alignment is a power of 2 allowing us to use ALIGNDOWN for rounding */
+	assert((delete_alignment & (delete_alignment - 1)) == 0);
+
+	if ((delete_zero_buf = malloc((size_t)delete_alignment)) == NULL){
```
ijc: cast should become unnecessary

src/lib/block_if.c
```diff
@@ -655,7 +698,33 @@ blockif_open(const char *optstr, const char *ident)
...
+		perror("Failed to allocate zeroed buffer for unaligned deletes");
+		goto err;
+	}
+	bzero(delete_zero_buf, (size_t)delete_alignment);
```
ijc: cast should become unnecessary

src/lib/block_if.c
ijc: Since we are using c99 I think `len_to_zero` and `written` can both be defined at the point of use (in both cases twice, in two disconnected scopes). e.g. you can write:
```
size_t len_to_zero = aligned_offset - arg.fp_offset;
...
ssize_t written = pwrite(bc->bc_fd, bc->bc_delete_zero_buf, len_to_zero, arg.fp_offset);
```
for the trim at the start.

djs55:
BTW I couldn't find a public link for the `fcntl` `F_PUNCHHOLE` API unfortunately. The relevant bits of the header seem to be:
```
#define F_PUNCHHOLE     99              /* Deallocate a range of the file */
...
/* fpunchhole_t used by F_PUNCHHOLE */
typedef struct fpunchhole {
        unsigned int fp_flags; /* unused */
        unsigned int reserved; /* (to maintain 8-byte alignment) */
        off_t fp_offset; /* IN: start of the region */
        off_t fp_length; /* IN: size of the region */
} fpunchhole_t;
```

I didn't realise `off_t` was signed. I added `assert`s to check they're >= 0 and then cast them to `size_t` for the rest of the calculations. I had to cast them back to signed `off_t` for the `pwrite` and `fcntl`, but they should still be +ve quantities. WDYT?

ijc:
Thanks, I found https://github.com/apple/darwin-xnu/blob/master/bsd/sys/fcntl.h which said the same thing.

Perhaps my advice should have been to use `off_t` instead, but I suspect that would mean a similar number of casts the other way. The `size_t` vs `off_t` dichotomy is always a bit annoying.

If we were being aggressively cautious as well as the negative assert you might also want to assert it is less than `SSIZE_MAX` (from `limits.h` IIRC) since this ends up being the upper limit of the `count` argument to `pwrite` which it wants to return. A pretty unlikely event IMHO so: LGTM.

PR #161, opened 2017-10-20
Port fwctl over from bhyve

ijc:
This enables the UEFI firmware to correctly detect the number of vcpus. Without this the guest only sees 1 vcpu regardless of the use of `-c` because the firmware tables (ACPI) won't be setup correctly.

The first commit simply imports the pristine files from FreeBSD, the second commit is the interesting one I guess.


MagnusS:
LGTM too - tested the binary and it solves the issue

PR #162, opened 2017-11-13
go: don't tie the users to golang's log

akimd:
- define a Logger interface to gain independance.
- really use the user's logger

If preferred, I can split this in three steps.

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -27,7 +27,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"log"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


rn:
I don't think we want logrus in here and if we do it needs to be vendored.

akimd:
The CI failure seems to be unrelated.

djs55:
At the moment the CI is using the homebrew version of `ocaml`, which occasionally updates and exposes some breakage-- I'm fixing these problems upstream atm. I think the `was built for newer OSX version` errors are coming from the binary libs shipped with the homebrew package too.

The ideal fix for both of these issues would be to pin the package versions and build everything (including the OCaml compiler) from source -- this is what we do in `moby/vpnkit`. I'll take a look at importing the scripts from there and see if I can make use of the CI cache to speed things up.

djs55:
@akimd to fix the CI, try rebasing on top of latest master.

akimd:
@djs55 Thanks, it worked!

djs55:
The logging change LGTM.

akimd:
Thanks.  But I'm not allowed to merge.

PR #163, opened 2017-11-14
Go: support exec.Cmd's extra files

akimd:
A bunch of minor stylistic changes, and a new feature, needed to allow Docker 4 Mac to rely on this hyperkit.go to launch hyperkit.

PR #164, opened 2017-11-15
ocaml: pin the versions of the compiler and libraries

djs55:
Unfortunately the OCaml compiler version in homebrew changes from time to time, as do the latest and greatest upstream packages. This PR pins the versions of the OCaml compiler and the dependencies to avoid failures like that in #163, using the same logic as moby/vpnkit.

By building the OCaml compiler from source, it should fix the linking warnings (and possibly) errors seen when running the artefact on older versions of macOS.


djs55:
CI is green again! (Hopefully more permanently this time)

PR #165, opened 2017-11-16
go: provide a means to change argv[0]

akimd:
In Docker for Mac, this is used so that all the processes belong to the `com.docker.*` family.


akimd:
Could you have a look, @djs55?  TIA!

djs55:
Looks good to me!

akimd:
Can't install it myself.

PR #166, opened 2017-11-20
Changes for Docker for Mac

akimd:
Currently there is some overlap between what hyperkit and piñata do, especially wrt the management of disk images.  Piñata can build raw images, and so does hyperkit.  Currently the code is duplicated, and Piñata will be adapted to use hyperkit on this area.

However Piñata also support qcow images, which requires `qcow-tool`, which is a dependency that I don't think hyperkit wants.  So the plan is to introduce an interface for disks in hyperkit, move the current support for raw images to comply with it, and then to introduce support for qcow images in Piñata, via this interface.

This is the plan.  This series of patches is meant to checkpoint: let's first have these first steps validated, then I'll proceed on both sides.

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -247,7 +247,6 @@ func (h *HyperKit) Start(cmdline string) error {
 }
 
 func (h *HyperKit) execute(cmdline string) error {
-	var err error
 	h.log.Printf("hyperkit: execute %#v", h)
 	// Sanity checks on configuration
 	if h.Console == ConsoleFile && h.StateDir == "" {
@@ -257,7 +256,7 @@ func (h *HyperKit) execute(cmdline string) error {
 		return fmt.Errorf("If ConsoleStdio is set but stdio is not a terminal, StateDir must be specified")
 	}
 	for _, image := range h.ISOImages {
-		if _, err = os.Stat(image); os.IsNotExist(err) {
+		if _, err := os.Stat(image); os.IsNotExist(err) {
 			return fmt.Errorf("ISO %s does not exist", image)
 		}
 	}
@@ -268,14 +267,14 @@ func (h *HyperKit) execute(cmdline string) error {
 		return fmt.Errorf("To forward vsock ports vsock must be enabled")
 	}
 	if h.Bootrom == "" {
-		if _, err = os.Stat(h.Kernel); os.IsNotExist(err) {
+		if _, err := os.Stat(h.Kernel); os.IsNotExist(err) {
 			return fmt.Errorf("Kernel %s does not exist", h.Kernel)
 		}
-		if _, err = os.Stat(h.Initrd); os.IsNotExist(err) {
+		if _, err := os.Stat(h.Initrd); os.IsNotExist(err) {
 			return fmt.Errorf("initrd %s does not exist", h.Initrd)
 		}
 	} else {
-		if _, err = os.Stat(h.Bootrom); os.IsNotExist(err) {
+		if _, err := os.Stat(h.Bootrom); os.IsNotExist(err) {
 			return fmt.Errorf("Bootrom %s does not exist", h.Bootrom)
 		}
 	}
@@ -306,10 +305,9 @@ func (h *HyperKit) execute(cmdline string) error {
 			h.Disks[idx] = config
 		}
 		if !strings.HasPrefix(config.Path, "file://") {
-			if _, err = os.Stat(config.Path); os.IsNotExist(err) {
+			if _, err := os.Stat(config.Path); os.IsNotExist(err) {
 				if config.Size != 0 {
-					err = CreateDiskImage(config.Path, config.Size)
-					if err != nil {
+					if err := CreateDiskImage(config.Path, config.Size); err != nil {
 						return err
 					}
 				} else {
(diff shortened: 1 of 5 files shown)


ijc:
I don't think it would be so bad to support qcow if the tool was available, so long as the error handling when the tool isn't there makes it clear what is going on. If someone has built Hyperkit with qcos2 support then they've also, at some point, built the ocaml qcow2 library which produces the tool too, so there is some chance it will be available.

One quirk to watch out for is that the go bindings also need to cope with the fact that the Hyperkit they are launching may lack qcow2 support even if the tool is available, hopefully Hyperkit itself produces a reasonable error message in that case. IOW the Go side should not have a compile time switch, just a runtime one.

akimd:
Even better then!  Thanks!

justincormack:
Please don't put random style changes into a PR that is about disk support.

akimd:
@justincormack I removed the Makefile bit.

akimd:
I doubt the failure at https://circleci.com/gh/moby/hyperkit/407?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link is due to my changes.

go/hyperkit.go
```diff
@@ -108,9 +110,11 @@ type HyperKit struct {
 
 	// VSock enables the virtio-socket device and exposes it on the host.
 	VSock bool `json:"vsock"`
+	// VSockDir specifies where the VSock will be.  Stands for StateDir if empty.
+	VSockDir string `json:"vsock_dir"`
 	// VSockPorts is a list of guest VSock ports that should be exposed as sockets on the host.
 	VSockPorts []int `json:"vsock_ports"`
-	// VSock guest CID
+	// VSock guest CID.
```
rn: please don't mix random style changes with like this with actual changes described in the commit message. Doing this makes it harder to review the actual changes.

go/hyperkit.go
```diff
@@ -108,9 +110,11 @@ type HyperKit struct {
 
 	// VSock enables the virtio-socket device and exposes it on the host.
 	VSock bool `json:"vsock"`
+	// VSockDir specifies where the VSock will be.  Stands for StateDir if empty.
```
rn: maybe "VSockDir specifies where the unix domain sockets will be created. Defaults to StateDir when empty."

go/hyperkit.go
```diff
@@ -467,10 +472,7 @@ func (h *HyperKit) buildArgs(cmdline string) {
 
 	for _, p := range h.Disks {
 		// Default the driver to virtio-blk
-		driver := "virtio-blk"
```
rn: This change also is completely unrelated to the change described in the commit message

go/hyperkit.go
```diff
@@ -248,6 +248,7 @@ func (h *HyperKit) Start(cmdline string) error {
 
 func (h *HyperKit) execute(cmdline string) error {
 	var err error
+	h.log.Printf("hyperkit: execute %#v", h)
```
rn: not sure how to comment on commit messages, but it should read something like "go: Improve logging" or some such. I read "add logs" and didn't know what it meant.

go/hyperkit.go
akimd: Well, it uses a function introduced in this commit.  I anticipated that someone might claim that a single-use function is useless.

go/hyperkit.go
```diff
@@ -305,14 +305,16 @@ func (h *HyperKit) execute(cmdline string) error {
 			config.Path = filepath.Clean(filepath.Join(h.StateDir, fmt.Sprintf("disk%02d.img", idx)))
 			h.Disks[idx] = config
 		}
-		if _, err = os.Stat(config.Path); os.IsNotExist(err) {
-			if config.Size != 0 {
-				err = CreateDiskImage(config.Path, config.Size)
-				if err != nil {
-					return err
+		if !strings.HasPrefix(config.Path, "file://") {
```
rn: I don't understand this change nor the commit message

akimd: `qcow-tool` wants the file to be specified as `"file:///this/is/my/image.qcow?sync=1&buffered=1"`.  Currently Docker for Mac passes this as "the path" since hyperkit.go offers no support for qcow, and the string must be passed as is to the driver.

So, as a transitional step until proper qcow support is implemented in hyperkit.go, we just need the existing check to be ignored for these files, as docker for mac already took care of this.

This will be removed, of course.

go/hyperkit.go
akimd: @rn Would you prefer that I rewrite the commits, or that I add fixes as additional commits?

go/hyperkit.go
rn: something with this sort of detail would be good to have in the commit message.

akimd: Indeed.

rn:
I'm very tempted to close the this PR. It has 10 commits sofar, some of the commits contain unrelated changes (including some really nit-picky ones adding a period at the end of comments) and the commits themselves touch different aspects of the go bindings.

I would prefer a PR per topic change, e.g. one which cleans up the VSock handling, one which improves the disk handling etc. That way I can actually look both at the entire diff and the individual commits without being distracted by too much noise. I would also suggest to make any style changes in a separate PR either as the first or the last one.

go/hyperkit.go
rn: This can still be done in a separate commit. You could mention in the commit which introduces the function as a single use function that subsequent commits will use it too. It's just as this commit stands now it changes too many different unrelated things

akimd: Ack.

akimd:
I also agree separate PR are better.  But it took a lot of time for my short PR last time, so I believed you guys preferred a single batch.  I was wrong.  I'll start again with several PRs, one at a time.

PR #167, opened 2017-11-21
Cleanup

akimd:
Stylistic changes.

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -289,24 +289,24 @@ func (h *HyperKit) execute(cmdline string) error {
 		}
 	}
 
-	for idx, config := range h.Disks {
-		if config.Path == "" {
+	for idx, disk := range h.Disks {
+		if disk.Path == "" {
 			if h.StateDir == "" {
 				return fmt.Errorf("Unable to create disk image when neither path nor state dir is set")
 			}
-			if config.Size <= 0 {
+			if disk.Size <= 0 {
 				return fmt.Errorf("Unable to create disk image when size is 0 or not set")
 			}
-			config.Path = filepath.Clean(filepath.Join(h.StateDir, fmt.Sprintf("disk%02d.img", idx)))
-			h.Disks[idx] = config
+			disk.Path = filepath.Clean(filepath.Join(h.StateDir, fmt.Sprintf("disk%02d.img", idx)))
+			h.Disks[idx] = disk
 		}
-		if _, err := os.Stat(config.Path); os.IsNotExist(err) {
-			if config.Size != 0 {
-				if err := CreateDiskImage(config.Path, config.Size); err != nil {
+		if _, err := os.Stat(disk.Path); os.IsNotExist(err) {
+			if disk.Size != 0 {
+				if err := CreateDiskImage(disk.Path, disk.Size); err != nil {
 					return err
 				}
 			} else {
-				return fmt.Errorf("Disk image %s not found and unable to create it as size is not specified", config.Path)
+				return fmt.Errorf("Disk image %s not found and unable to create it as size is not specified", disk.Path)
 			}
 		}
 	}
@@ -353,8 +353,8 @@ func (h *HyperKit) IsRunning() bool {
 
 // isDisk checks if the specified path is used as a disk image
 func (h *HyperKit) isDisk(path string) bool {
-	for _, config := range h.Disks {
-		if filepath.Clean(path) == filepath.Clean(config.Path) {
+	for _, disk := range h.Disks {
+		if filepath.Clean(path) == filepath.Clean(disk.Path) {
 			return true
 		}
 	}
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -12,9 +12,7 @@
 // there.
 //
 // Currently this module has some limitations:
-// - Only supports zero or one disk image
 // - Only support zero or one network interface connected to VPNKit
-// - Only kexec boot
 //
 // This package is currently implemented by shelling out a hyperkit
 // process. In the future we may change this to become a wrapper
@@ -41,9 +39,9 @@ import (
 )
 
 const (
-	// ConsoleStdio configures console to use Stdio
+	// ConsoleStdio configures console to use Stdio.
 	ConsoleStdio = iota
-	// ConsoleFile configures console to a tty and output to a file
+	// ConsoleFile configures console to a tty and output to a file.
 	ConsoleFile
 
 	defaultVPNKitSock = "Library/Containers/com.docker.docker/Data/s50"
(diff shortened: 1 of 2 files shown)


go/doc.go
```diff
@@ -0,0 +1,22 @@
+// Package hyperkit provides a Go wrapper around the hyperkit
```
ijc: > As is customary in go when a package has several files.

There are three files here `hyperkit.go` and `pty_util_{darwin,fallback}.go`. Keeping the docs in `hyperkit.go` makes sense given the peripheral nature of the other files. Please drop this commit.

go/hyperkit.go
```diff
@@ -675,3 +672,13 @@ func getHome() string {
...
+
+// defaultString returns the first non empty string, or "" if there is none.
+func defaultString(ss ...string) string {
+	for _, s := range ss {
+		if s != "" {
+			return s
+		}
+	}
+	return ""
+}
```
ijc: This is not a simplification over the existing code, please leave the code as it was.

go/hyperkit.go
```diff
@@ -298,12 +298,7 @@ func (h *HyperKit) execute(cmdline string) error {
 
 	// Run
 	h.buildArgs(cmdline)
-	err = h.execHyperKit()
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return h.execHyperKit()
```
ijc: I prefer the style in which all the actions+checks are written in the same manner:
```
if !something {
    return error
}
[... more...]
if !something_else {
    return error
}
return success
```
i.e. to not special case the final one as you are doing here (and in the other similar places). It makes the flow more consistent and reduces churn if you need to add a new thing after the final one.

Changing to the `if err := foo ; err != nil` style for these would be fine though.

go/hyperkit.go
```diff
@@ -376,7 +360,7 @@ func (h *HyperKit) Remove(keepDisk bool) error {
 	return nil
 }
 
-// Convert to json string
+// Convert to json string.
```
ijc: This is in the wrong patch

go/doc.go
akimd: Hi @ijc.  Thanks for the review.

Okay.  I'm also creating more files, with log.go and disk.go, and `doc.go` is the common practice I saw.  But I don't care and sure can drop that.

go/hyperkit.go
akimd: I concur on this, and on many regards I dislike the coding style in Go.

However, there is one good thing in Go, and it is that they tend to enforce one way to write things, enforced by tools such as golint etc.  And in that case the style that you and I prefer is not the standard one.

So for instance with
```go
	if err := h.execHyperKit(); err != nil {
		return err
	}

	return nil
```
I get
```
$ golint
hyperkit.go:317:2: redundant if ...; err != nil check, just return error instead.
```
So I would still claim this commit is in the right direction.

go/hyperkit.go
akimd: The function will be used elsewhere, that's not its single use.

go/hyperkit.go
akimd: Ouch.  Good catch, thanks!

go/hyperkit.go
rn: Then it should go into the PR which does this. It would also be good to mention it in the commit message as i proposed here https://github.com/moby/hyperkit/pull/166/commits/c202f0709cbc834972a91f58990875f1d0a34ac7#r152323406

akimd: That's probably not the link you meant to sent, but rather https://github.com/moby/hyperkit/pull/166/commits/c202f0709cbc834972a91f58990875f1d0a34ac7.  But maybe GitHub is playing tricks on URLs here.

I'll move it into another PR.

go/hyperkit.go
ijc: I disagree with `golint` on this matter, normally I just have to stomach it but since I'm a maintainer on this project I don't have to in this case ;-). If the other maintainers disagree I'm happy to be overruled by them.

NB also this paragraph from https://github.com/golang/lint:
> The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process. Golint makes suggestions for many of the mechanically checkable items listed in Effective Go and the CodeReviewComments wiki page.

In this particular case I believe the (usually good) recommendation to avoid conditionals which don't actually impact program flow does not always apply at the end of a function.

rn: I'm fine either way

akimd: @ijc Yes, sure, golint guys _had_ to write that.  It's like warnings in C/C++, they're just that, warnings.  However all the Go projects I've seen so far (but I'm new to this language, granted) pay attention to go fmt, go vet, golint.

I'll remove this commit entirely.

akimd:
Well, now the CI dies because it uses golint.  Should I teach it to ignore the diagnostics?

rn:
well, one option would be to not do the refactor and just add the functionality you need. From what I remember it didn't rely on the refactoring

ijc:
Yuk, seems it cares about `if err := ... ; err != nil` but not
```
err = ...
if err != nil {
```
I do prefer the `if err :=` variant. So it seems the choices are:
1. accept the `err =  ... \n if err != ` style (i.e. drop the refactor here which then triggers the lint of the following item)
2. accept the special casing of the final return
3. somehow grep out lint warnings we don't approve of
4. don't run golint at all.

Looks like e.g. containerd use gometalinter which makes it easier to achieve 3 by providing a way to exclude output by giving regexps.

What do other maintainers think? Using gometalinter might be a nice improvement in its own right at some point, but I think given how long this PR has been ongoing 1 is probably the right approach for now (and it seems @rn at least feels the same).

akimd:
Well, guys, I will need your guidance here.  I clearly messed up something.

I read about everything in hyperkit.go in order to understand it, and to be ready to contribute to it.  While reading I noted, and addressed, some issues (minor, granted) such as comments not as typically recommended in go, large variable scopes (I'm already seen code -elsewhere- treating the wrong `err` because the previous one leaked.  I do feel strongly against large variable scopes, and I might be psychotic on this regard), duplicate code, imperative code where even C would be better looking thanks to `?:`, write tests, etc.

IOW, I tried to give as much love as I could to this piece of code.  Just as I would, and did, in the projects I contribute to.  And it started because I felt that both d4m and hyperkit.go had similar code/concerns, but d4m provides more features, so it would better to move it to hyperkit.go.

I clearly was too lazy at the beginning by leaving unrelated things together (a bad habit I took recently in D4M because we just don't have/take enough time to slice the commits as they should), but I think I addressed these concerns.

We reached a point where I feel we are all wasting time on this.

What exactly are your expectations for contributions?

ijc:
I think ultimately the go bindings here are a very small piece of code which does not require a lot of development or maintenance and which have not been subject to very much churn in the past. The changes made here are a reasonably large amount of churn in the context this project.

I'll also note that other project (such as the Linux kernel) discourage pure "coding style" changes and prefer instead to see things fixed incrementally as they are changed for other reasons. I think we are coming to the realisation (rather too late for this PR, for which I apologise), that this model is the one we prefer for this project.

I propose we put this PR to one side and instead look towards the functional changes you would like to make, and restrict the cleanups to those which are directly related to those.

The `gometalint` stuff is something I think we should perhaps look at, but not necessarily in the short term.

akimd:
Hi Ian,

Ok, thanks for the detailed answer.

The commits that do contribute changes are on top of the aforementioned stylistic ones.  Currently I'm on other issues, I'll address the conflicts caused by their removal later.

Cheers.

PR #168, opened 2017-12-02
File URIs are now valid inputs for DiskConfig paths

dprotaso:
This is needed to support qcow2 disks given that
file uris seem to be required.

When using a qcow2 disk with a non-uri path you'll get
the error

"Config.to_string expected a string of the form file://<path>?sync=(none|os|drive)&buffered=(0|1)"

Signed-off-by: dave <dprotaso@gmail.com>

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -28,6 +28,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net"
+	"net/url"
 	"os"
 	"os/exec"
 	"os/user"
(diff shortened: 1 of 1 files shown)


go/hyperkit.go
```diff
@@ -299,9 +300,18 @@ func (h *HyperKit) execute(cmdline string) error {
 			config.Path = filepath.Clean(filepath.Join(h.StateDir, fmt.Sprintf("disk%02d.img", idx)))
 			h.Disks[idx] = config
 		}
-		if _, err = os.Stat(config.Path); os.IsNotExist(err) {
+
+		// assume the path could be a file:// uri with query params
+		// use the actual path from the parsed result
+		u, err := url.Parse(config.Path)
+
```
rn: Super minor nit, but there shouldn't be a empty line here

dprotaso:
nit is gone

PR #172, opened 2017-12-06
Changes for Docker for Mac

akimd:
This is the changes that are currently running in Docker for Mac.  I have removed the stylistic issues (adjusted comments, scope reduction, etc.) that were discussed in https://github.com/moby/hyperkit/pull/167.  The `defaultString` function is removed at the end.

Cheers!

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -30,6 +30,7 @@ build/hyperkitgo.exe: $(DEPS)
 	go build -o $@ --ldflags '-extldflags "-fno-PIC"' \
 		sample/main.go
 
+.PHONY: clean fmt lint check
 clean:
 	rm -rf build
 
@@ -42,6 +43,10 @@ lint:
 	$(if $(shell which golint || echo ''), , \
 		$(error Please install golint))
 	@test -z "$$(golint ./... 2>&1 | grep -v ^vendor/ | grep -v mock/ | tee /dev/stderr)"
+
+check:
+	@echo "+ $@"
+	@go test . $(TESTFLAGS)
 
 # this will blow away the vendored code and update it to latest
 .PHONY: vendor vendor-local
--- a/go/Makefile
+++ b/go/Makefile
@@ -30,7 +30,7 @@ build/hyperkitgo.exe: $(DEPS)
 	go build -o $@ --ldflags '-extldflags "-fno-PIC"' \
 		sample/main.go
 
-.PHONY: clean fmt lint check
+.PHONY: clean fmt lint test
 clean:
 	rm -rf build
 
@@ -44,9 +44,13 @@ lint:
 		$(error Please install golint))
 	@test -z "$$(golint ./... 2>&1 | grep -v ^vendor/ | grep -v mock/ | tee /dev/stderr)"
 
-check:
+test:
 	@echo "+ $@"
-	@go test . $(TESTFLAGS)
+# We pass the module path (not just `.`) because in the CI we play
+# tricks with symlinks that confuse go test: when we enter `go/`, we
+# are no longer in the go tree, but back in ~/distiller/hyperkit/go
+# where Circle CI forces our checkout.
+	@go test -v $(TESTFLAGS) github.com/moby/hyperkit/go
 
 # this will blow away the vendored code and update it to latest
 .PHONY: vendor vendor-local
@@ -67,6 +71,6 @@ vendor-local:
 setup:
 	go get github.com/golang/lint/golint
 
-ci: setup build/hyperkitgo build/hyperkitgo_linux build/hyperkitgo.exe
+ci: setup build/hyperkitgo build/hyperkitgo_linux build/hyperkitgo.exe test
 	test -z "$$(gofmt -s -l . 2>&1 | grep -v ^vendor/ | tee /dev/stderr)"
 	test -z "$$(golint ./... 2>&1 | grep -v ^vendor/ | grep -v mock/ | tee /dev/stderr)"
--- a/go/disk.go
+++ b/go/disk.go
@@ -243,6 +243,13 @@ func (d *QcowDisk) QcowTool(verb string, args ...string) *exec.Cmd {
 	return exec.Command(path, append([]string{verb, d.Path}, args...)...)
 }
 
+func run(cmd *exec.Cmd) (string, error) {
+	buf, err := cmd.CombinedOutput()
+	out := string(buf)
+	log.Debugf("ran %v: out=%q, err=%v", cmd.Args, out, err)
+	return out, err
+}
+
 // Exists iff the image file can be stat'd without error.
 func (d *QcowDisk) Exists() bool {
 	return exists(d)
@@ -255,11 +262,8 @@ func (d *QcowDisk) Ensure() error {
 
 // Create a disk with the given size in MiB
 func (d *QcowDisk) create() error {
-	cmd := d.QcowTool("create", "--size", fmt.Sprintf("%dMiB", d.Size))
-	if err := cmd.Start(); err != nil {
-		return err
-	}
-	return cmd.Wait()
+	_, err := run(d.QcowTool("create", "--size", fmt.Sprintf("%dMiB", d.Size)))
+	return err
 }
 
 // The current disk size in MiB.
@@ -267,8 +271,10 @@ func (d *QcowDisk) getCurrentSize() (int, error) {
 	if _, err := os.Stat(d.Path); err != nil {
 		return 0, err
 	}
-	cmd := d.QcowTool("info", "--filter", ".size")
-	out, err := cmd.Output()
+	out, err := run(d.QcowTool("info", "--filter", ".size"))
+	if err != nil {
+		return 0, err
+	}
 	size, err := strconv.ParseInt(strings.TrimSpace(string(out)), 10, 64)
 	if err != nil {
 		return 0, err
@@ -278,8 +284,8 @@ func (d *QcowDisk) getCurrentSize() (int, error) {
 
 // Resize the virtual size of the disk
 func (d *QcowDisk) resize() error {
-	cmd := d.QcowTool("resize", "--size", fmt.Sprintf("%dMiB", d.Size))
-	return cmd.Run()
+	_, err := run(d.QcowTool("resize", "--size", fmt.Sprintf("%dMiB", d.Size)))
+	return err
 }
 
 func (d *QcowDisk) sizeString() string {
@@ -296,7 +302,7 @@ func (d *QcowDisk) compact(lockFh *os.File) error {
 	cmd := d.QcowTool("compact")
 	// Pass a reference to the file lock to the compact program
 	cmd.ExtraFiles = []*os.File{lockFh}
-	if err := cmd.Run(); err != nil {
+	if _, err := run(cmd); err != nil {
 		if err.(*exec.ExitError) != nil {
 			return errors.New("Failed to compact qcow2")
 		}
@@ -311,7 +317,7 @@ func (d *QcowDisk) check(lockFile *os.File) error {
 	cmd := d.QcowTool("check")
 	// Pass a reference to the file lock to the compact program.
 	cmd.ExtraFiles = []*os.File{lockFile}
-	if err := cmd.Run(); err != nil {
+	if _, err := run(cmd); err != nil {
 		if err.(*exec.ExitError) != nil {
 			return errors.New("qcow2 failed integrity check: it may be corrupt")
 		}
--- a/go/disk.go
+++ b/go/disk.go
@@ -233,8 +233,11 @@ func (d *QcowDisk) String() string {
 
 // QcowTool prepares a call to qcow-tool on this image.
 func (d *QcowDisk) QcowTool(verb string, args ...string) *exec.Cmd {
-	return exec.Command(defaultString(d.QcowToolPath, "qcow-tool"),
-		append([]string{verb, d.Path}, args...)...)
+	path := d.QcowToolPath
+	if path == "" {
+		path = "qcow-tool"
+	}
+	return exec.Command(path, append([]string{verb, d.Path}, args...)...)
 }
 
 // Exists if the image file exists.
@@ -333,7 +336,13 @@ func (d *QcowDisk) Stop(lockFile *os.File) error {
 // AsArgument returns the command-line option to pass after `-s <slot>:0,` to hyperkit for this disk.
 func (d *QcowDisk) AsArgument() string {
 	res := fmt.Sprintf("%s,file://%s?sync=%s&buffered=1", diskDriver(d.Trim), d.Path, d.OnFlush)
-	res += fmt.Sprintf(",format=%v", defaultString(d.Format, "qcow"))
+	{
+		format := d.Format
+		if format == "" {
+			format = "qcow"
+		}
+		res += fmt.Sprintf(",format=%v", format)
+	}
 	if d.Stats != "" {
 		res += ",qcow-stats-config=" + d.Stats
 	}
--- a/go/disk.go
+++ b/go/disk.go
@@ -3,10 +3,106 @@ package hyperkit
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 )
 
-// RawDisk describes a disk image.
+const (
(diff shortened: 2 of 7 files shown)


go/disk.go
```diff
@@ -0,0 +1,352 @@
...
+
+// compact the disk to shrink the physical size.
+func (d *QcowDisk) compact(lockFh *os.File) error {
+	log.Infof("Compact: %q... (%v)", d, d.sizeString())
+	cmd := d.QcowTool("compact")
+	// Pass a reference to the file lock to the compact program
+	cmd.ExtraFiles = []*os.File{lockFh}
+	if err := cmd.Run(); err != nil {
+		if err.(*exec.ExitError) != nil {
+			return errors.New("Failed to compact qcow2")
```
ijc: `d.QcowTool` just calls `exec.Command` which I don't think does anything with `cmd.Stdout` (or `Stderr`) by default, meaning they get redirected to `/dev/null`. I think we probably want the error messages from the tool to go somewhere.

Looking at the callers of `QcowTool` I think they might all be happy with simply directing the output to Stdio, i.e. `QcowTool` could set `Stdout` and `Stderr` to `os.Stdout` and `os.Stderr` respectively. If you want to put the output somewhere other than stdio (to the logs perhaps) then you would want to use `cmd.CombinedOutput` (I think that is better that `cmd.Output` for the cases here) and log the output in the error case (I think in all the users not just this one).

go/disk.go
```diff
@@ -0,0 +1,352 @@
...
+	s, err := d.getFileSize()
+	if err != nil {
+		return fmt.Sprintf("cannot get size: %v", err)
+	}
+	return fmt.Sprintf("%vMiB", s)
+}
+
+// compact the disk to shrink the physical size.
+func (d *QcowDisk) compact(lockFh *os.File) error {
+	log.Infof("Compact: %q... (%v)", d, d.sizeString())
```
ijc: You format strings with `%v` quite a lot in this PR, is there any reason to prefer `%v` over `%s`? If not (and I don't think there is, but perhaps I'm missing something) please use `%s`.

go/hyperkit.go
```diff
@@ -474,24 +425,17 @@ func (h *HyperKit) buildArgs(cmdline string) {
...
+		a = append(a, "-s", fmt.Sprintf("%d:0,%s", nextSlot, disk.AsArgument()))
 		nextSlot++
 	}
 
 	if h.VSock {
-		l := fmt.Sprintf("%d,virtio-sock,guest_cid=%d,path=%s", nextSlot, h.VSockGuestCID, h.StateDir)
+		path := h.VSockDir
+		if path == "" {
+			path = h.StateDir
+		}
```
ijc: My preference would be to arrange that `h.VSockDir` is always set by the time we get to this point, by including in `execute` (I think, maybe there is somewhere better) after the sanity checks a:
```
if h.VSockDir == "" {
    h.VsockDir = h.StateDir
}
```
Probably right before `// Create files`.

go/disk.go
```diff
@@ -0,0 +1,352 @@
...
+
+	create() error
+	getFileSize() (int, error)
+	resize() error
+}
+
+// exists if the image file exists.
+func exists(d Disk) bool {
+	_, err := os.Stat(d.GetPath())
+	return err == nil
```
ijc: `os.Stat` can return errors other than `os.ErrNotExist`.

Should either log in that case or return the `err` for the caller to deal with, which would be something like:
```
if err== nil || os.IsNotExist(err) {
    return err == nil, nil
}
return false, err
```
Perhaps that can be simplified though.

I can't actually find any callers of the `Exists` functions which wrap this in the PR though so I'm not sure which would be more appropriate. Maybe this could even just be removed if it is unused.

go/hyperkit.go
```diff
@@ -281,43 +264,29 @@ func (h *HyperKit) execute(cmdline string) error {
...
-			config.Path = filepath.Clean(filepath.Join(h.StateDir, fmt.Sprintf("disk%02d.img", idx)))
-			h.Disks[idx] = config
+			disk.SetPath(filepath.Clean(filepath.Join(h.StateDir, fmt.Sprintf("disk%02d.img", idx))))
+			h.Disks[idx] = disk
 		}
-
-		// assume the path could be a file:// uri with query params
-		// use the actual path from the parsed result
-		u, err := url.Parse(config.Path)
-		if err != nil {
```
ijc: This does not seem to have moved into the `RawDisk` handler. Is that an error introduced by rebasing over #168 or something deliberate? I think this is going to have an effect on the `linuxkit` tool's use of this interface, which currently allows the user to simply pass in the path (which might be a `file:///path` thing AIUI).

I don't think we want to teach `linuxkit` to parse those URLs in order to construct a `QcowDisk` object, or add a lot of options to the CLI so users can specify all the options independently in order to construct it, so I think we need to continue to support `RawDisk` where the `.Path` is structured as a URL still. I suspect (but I'm not sure) that this is just a case of replacing the `url.Parse` in the `Exists()` function.

go/disk.go
```diff
@@ -0,0 +1,352 @@
...
+// > Unfortunately the `virtio-blk` protocol doesn't support `TRIM`
+// > requests at all so we have to use `ahci-hd` (if you try to run
+// > `fstrim /var/lib/docker` with `virtio-blk` it'll give an `ioctl`
+// > error).
+func diskDriver(trim bool) string {
+	if trim {
+		return "ahci-hd"
+	}
+	return "virtio-blk"
+}
```
ijc: I notice that the user's ability to specify an override driver (by setting `Driver` in the config) has silently disappeared here. Is that deliberate? Perhaps it could be retained for the raw disk case?

go/disk.go
akimd: Hi Ian,

Yes, I am well aware of the different types of error.  But this is the code as it lived in Piñata, and over there, what mattered is whether the file existed.  Errors counted as "does not exist", and then the code that deals with the non-existing situation would deal with finer grained errors.

But I can definitely add logs here.

go/disk.go
akimd: We could add it, but as far as I could see from the code used before, the ability to specify the driver was meant only for this: computing the driver from trim.

go/disk.go
```diff
@@ -0,0 +1,352 @@
...
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	return cmd.Wait()
+}
+
+// Query the current virtual size of the disk in MiB
+func (d *QcowDisk) getFileSize() (int, error) {
+	// Return a failure if the file doesn't exist yet
+	if _, err := os.Stat(d.Path); err != nil {
```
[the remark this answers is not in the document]
akimd: I'll remove the comment.

go/disk.go
akimd: Well, this is Go, not C.  We have decent variadic functions that do not require the users to do the compiler's job of tracking types.  Boost.Format, Lisp, Python, Rust, etc. all go away from the %d/%s model.  Go's universal format spec is %v.

What's to be preferred with %s?

go/disk.go
akimd: Sounds great!  Will do that.

go/hyperkit.go
akimd: I had no realised that it was considered a _feature_ to have this.  The current code does not even checks that it's not say `https://` which is used.

What exactly should be the interface?

go/hyperkit.go
akimd: Will do.

go/hyperkit.go
ijc: `linuxkit run hyperkit -disk «foo»` needs to keep working for both raw and qcow disks.

This code was, AIUI, just checking that the path part was valid. Whether or not the schema was valid is not checked until hyperkit itself tries to use it, when it will probably reject `https` (unless @djs55's qcow library is cleverer than I imagined). This was only added earlier in the week in #168.

go/disk.go
ijc: OK, it doesn't look like the `linuxkit` CLI used this either, so lets drop, someone can always put it back later if its needed.

go/disk.go
ijc: Mainly that `%s` doesn't need to do a bunch of runtime introspection, probably doesn't matter for this case, lets leave it then.

go/Makefile
```diff
@@ -43,6 +44,10 @@ lint:
 		$(error Please install golint))
 	@test -z "$$(golint ./... 2>&1 | grep -v ^vendor/ | grep -v mock/ | tee /dev/stderr)"
 
+check:
```
ijc: Is this new since the first round of review or did I miss it somehow?

Anyway, there is `make test` in the top level , might be nice to be consistent with the naming, or possibly even recurse (not so sure about that, I think perhaps not, although a `test-go` might be handy). I think you probably do want to hook it up in the CI though, should be a simple to add a dep to the existing `ci` rule I suppose.

ijc:
Looks like the LinuxKit update is pretty simple:
```diff
diff --git a/src/cmd/linuxkit/run_hyperkit.go b/src/cmd/linuxkit/run_hyperkit.go
index 047c35406..34d810b08 100644
--- a/src/cmd/linuxkit/run_hyperkit.go
+++ b/src/cmd/linuxkit/run_hyperkit.go
@@ -204,7 +204,7 @@ func runHyperKit(args []string) {
 		if d.Path == "" {
 			log.Fatalf("disk specified with no size or name")
 		}
-		hd := hyperkit.DiskConfig{Path: d.Path, Size: d.Size}
+		hd := &hyperkit.RawDisk{Path: d.Path, Size: d.Size}
 		h.Disks = append(h.Disks, hd)
 	}

```
At least it compiled with just that and seems logically like it should be enough.

I think all which remains is the need to keep [`linuxkit run hyperkit -disk «foo»`](https://github.com/moby/hyperkit/pull/172#discussion_r155522176) working when someone tries to use a qcow disk.

go/disk.go
akimd: Good point!  But since it's all based on `interface{}`, it must do that anyway I guess.  It doesn't seem to make much of a difference.
```go
package foo

import (
	"fmt"
	"testing"
)

func BenchmarkVString(b *testing.B) {
	s := "Hello, world!"
	for i := 0; i < b.N; i++ {
		fmt.Sprintf("%v", s)
	}
}

func BenchmarkVInt(b *testing.B) {
	s := 42
	for i := 0; i < b.N; i++ {
		fmt.Sprintf("%v", s)
	}
}

func BenchmarkSString(b *testing.B) {
	s := "Hello, world!"
	for i := 0; i < b.N; i++ {
		fmt.Sprintf("%s", s)
	}
}

func BenchmarkDInt(b *testing.B) {
	s := 42
	for i := 0; i < b.N; i++ {
		fmt.Sprintf("%d", s)
	}
}
```
```
go test -bench .
goos: darwin
goarch: amd64
BenchmarkVString-8   	10000000	       109 ns/op
BenchmarkVInt-8      	20000000	        97.8 ns/op
BenchmarkSString-8   	10000000	       112 ns/op
BenchmarkDInt-8      	20000000	        97.7 ns/op
```

go/hyperkit.go
akimd: I am really reluctant to trying to force this URI into the Path.  The whole point of the interface was precisely to make this syntactic difference go away.

So am I correctly understanding that linuxkit's interface for disk expect either `foo.raw` or `file:///foo.qcow`?  Shouldn't it rather rely on conventions on the file extension?

I suppose the best option would be something like `MakeDisk` that hides the instantiation of RawDisk/QcowDisk, and I certainly agree it would be a nice addition, but it's its input which is still unclear to me.  If I could see a few concrete examples of `«foo»` in `linuxkit run hyperkit -disk «foo»`, I'll do that.  How you want to specify options for raw and for qcow would help me.

go/Makefile
akimd: I have added a bunch of commits, including this one which adds tests on qcow.

Indeed, I missed the `test` target in the top-level.  And I'm used to the GNU Coding Standards where this guy is named `check` :)

I'll rename it.

I'll leave the decision to recur to you though: currently the top level Makefile does not walk into `go/` at all.  But I'll look at the CI.

akimd:
WRT the linuxkit bit, I don't know what `d` is, but after all maybe it could be `RawDisk` directly?  It's a pity that controls are performed in several places: the check for `Path == ""` should be delegated/left to hyperkit.go.  WDYT?

go/disk.go
dprotaso: I was exposing it here - for the purposes of specifying ahci-hd for qcow & trimming
https://github.com/linuxkit/linuxkit/pull/2791

go/hyperkit.go
dprotaso: The intention of https://github.com/moby/hyperkit/pull/168 was to allow file:// uris to be passed to hyperkit. This was an expectation of hyperkit for qcow files

go/disk.go
akimd: Hi David!
Just to make sure: you're saying that computing the driver is sufficient?  Or do you need a means to specify it?

go/hyperkit.go
ijc: > So am I correctly understanding that linuxkit's interface for disk expect either foo.raw or file:///foo.qcow? Shouldn't it rather rely on conventions on the file extension?

It expects the same things the hyperkit CLI does I think since apart from deconstructing it to check the path element exists it just passes it through in (before you PR) `fmt.Sprintf("%d:0,%s,%s", nextSlot, driver, p.Path)`.

So assuming I've understood the CLI correctly (cc @djs55), yes, it expects `foo.raw` or `file:///foo.qcow` or `file://foo.qcow2?someoption=yes`.

ijc:
`d` here is a `linuxkit.DiskConfig` (the linuxkit code is open, please take a look). While `linuxkit` could parse the input, put it in a `QcowDisk` and have `hyperkit.go` reconstruct the input URL. That seems a bit silly to me and roundabout to me.

I think it is fine to have `QcowDisk` as a data structure driver way to have a qcow disk for callers that want it and `RawDisk` as the "passthrough" variant (so more like `Raw` in the sense of "not really understood or introspected very deeply by `hyperkit.go`" as opposed to the disk format).

I suppose we could split `RawDisk` into the "raw as in file format" and some third `FooDisk` type which represents the "don't look at this just pass it through" way which which doesn't do any checks (including the path check), but that seems unnecessary compared with just reinstating the URL parse in the RawDisk path which should be a three line fix on the current state of this PR.

go/Makefile
ijc: Re recursing please could you just add after the `test-qcow` line at the top level something like:
```
test-go:
    @make -C go test
```

Up to you whether you then want to have `circle.ci` then say `make go-test` or `make -C go test`, I don't mind.

akimd:
Hi Ian,

I think that what really was bothering me is that the file format (raw, qcow, whatever) is treated differently: genuine path for raw, and uri-like for qcow.  What would you think of supporting the two syntaxes for both formats?  Additional arguments for raw could be embedded in the URI just like for qcow, and qcow could use its default values with a plain path?  And since this format (which is also what you are describing AFAICT) is a black box for linuxkit, it would benefit directly from new formats.

Actually, even Size could be just `&size=42`.

That's where the NewDisk I mentioned kicks in, in replacement for `hd := &hyperkit.RawDisk{Path: d.Path, Size: d.Size}`.

WDYT?

ijc:
At this stage I would prefer to make just the minimal modifications to this rather long running PR to remove the regression for linuxkit's usage.

Any further improvements on top of that along the lines of what you are suggesting can be done in a follow up if there is appetite to do so.

akimd:
Ian, this should match your expectations.

akimd:
The failure is unrelated to my PR, but I can't rerun the job on CircleCI.

djs55:
I hit the rebuilt button.

322-334:
This is really aggravating for me. I have an ex-girlfriend who has been hijacking my Android phones for a long time now. Now she is somehow accessing the phones remotely. I'm just learning about all this stuff. How do I protect this phone. So far she is unaware of this phone, as i bought it just recently.  I'm not a tech and could use the advice.

PR #173, opened 2017-12-07
Lock the backing file to prevent concurrent modification

djs55:
Previously it was possible to launch `hyperkit` twice using the same backing file, leading to disk corruption.

This patch uses `flock` advisory locking to prevent 2 `hyperkit` instances clashing. Note since the lock is advisory it's still possible for other processes to concurrently modify the files.

For raw format files:
- if read-only: use `flock` with `LOCK_SH` to allow other `hyperkit` instances to open the file read-only but prevent any other `hyperkit` opening the file for read+write
- if read-write: use `flock` with `LOCK_EX` to prevent any other `hyperkit` instance from opening the file for reading or writing.

For qcow format files we request a lock via the new API in `mirage-block-unix.2.9.0`. This PR also includes an ocaml metadata update.

In all cases we call `flock` with `LOCK_NB` which means the call will not block; if someone else has the lock it will be treated as a permanent failure and reported to the user.

two hyperkit bears fighting over a fish representing a disk


akimd:
So once we update hyperkit in D4M, we can safely get rid of our own lock file, right?

djs55:
Yes, we can get rid of the other lock file and everything should be a bit
simpler :)

PR #174, opened 2017-12-15
Go: revamp the Start/Run procedures

akimd:
Currently we offer two means to run hyperkit: `Start` and `Run`. Both run the process in background (using `exec.Cmd.Start`), in order to be able to save the state when launching succeeded.

Both wait for the command to finish.  `Run` returns the exit status. `Start` waits in a goroutine and discard the result.

As a result, the user has no clean means to run in background and get the exit status.  The `Start` function discards the exit status.  But since `Run` is blocking, the caller must run it in a goroutine.  Therefore the callers needs to watch the process to know when it is actually running (e.g., by watching `h.Pid`).

Revamp this.
- First, make changes easier by isolating the checking and the `exec.Cmd` creation parts into their own functions
- Provide the HyperKit struct with a channel to publish the result of waiting for hyperkit
- Move what remains of `execute`/`execHyperKit` into `Start`
- Write `Run` on top of `Start`.

I have not included the `exec.Cmd` in the `HyperKit` struct, but that was an option.

Preparing the `Cmd` and leaving it to the user was also an option.  I stayed with the current approach.

Docker for Mac is made cleaner/faster thanks to these changes.
```diff
-   hypErr := make(chan error, 1)
-
-   go func() { hypErr <- h.Run("") }()
-
-   // Wait for the process to be created.
-   for i := 0; i < 10; i++ {
-           if h.Pid != 0 {
-                   break
-           }
-           select {
-           case err := <-hypErr:
-                   log.Fatalf("Hyperkit failed to start: %v", err)
-           default:
-                   log.Infof("Waiting for %#v to start", h)
-                   time.Sleep(500 * time.Millisecond)
-           }
-       }
-   if h.Pid == 0 {
-           log.Fatalf("hyperkit did not start")
-   }
-   return h.Pid, hypErr
+   if err := h.Start(""); err != nil {
+           log.Fatalf("Hyperkit failed to start: %v", err)
+       }
+   return h.Pid, h.ErrorCh
```

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -131,6 +131,9 @@ type HyperKit struct {
 	// process died unexpectedly.
 	ExtraFiles []*os.File `json:"extra_files"`
 
+	// ErrorCh is where the result of Wait'ing for hyperkit is sent.
+	ErrorCh chan error `json:"-"`
+
 	// Below here are internal members, but they are exported so
 	// that they are written to the state json file, if configured.
 
@@ -141,8 +144,7 @@ type HyperKit struct {
 	// CmdLine is a single string of the command line
 	CmdLine string `json:"cmdline"`
 
-	process    *os.Process
-	background bool
+	process *os.Process
 }
 
 // New creates a template config structure.
@@ -170,6 +172,8 @@ func New(hyperkit, vpnkitsock, statedir string) (*HyperKit, error) {
 	h.VSockGuestCID = defaultVSockGuestCID
 
 	h.Console = ConsoleStdio
+
+	h.ErrorCh = make(chan error, 1)
 
 	return &h, nil
 }
@@ -204,6 +208,8 @@ func FromState(statedir string) (*HyperKit, error) {
 		return nil, err
 	}
 
+	h.ErrorCh = make(chan error, 1)
+
 	return h, nil
 }
 
@@ -213,16 +219,33 @@ func (h *HyperKit) SetLogger(l Logger) {
 	SetLogger(l)
 }
 
-// Run the VM with a given command line until it exits
+// Run the VM with a given command line until it exits.
 func (h *HyperKit) Run(cmdline string) error {
-	h.background = false
-	return h.execute(cmdline)
+	if err := h.Start(cmdline); err != nil {
+		return err
+	}
+	return <-h.ErrorCh
 }
 
-// Start the VM with a given command line in the background
+// Start the VM with a given command line in the background.  Return
+// failures to start the process.  Process errors are published in
+// h.ErrorCh.
 func (h *HyperKit) Start(cmdline string) error {
-	h.background = true
-	return h.execute(cmdline)
+	log.Debugf("hyperkit: Start %#v", h)
+	if err := h.check(); err != nil {
+		return err
+	}
+	h.buildArgs(cmdline)
+	cmd, err := h.makeCommand()
+	if err != nil {
+		return err
+	}
+	// Make sure we reap the child when it exits
+	go func() {
+		log.Debugf("hyperkit: Waiting for %#v", cmd)
+		h.ErrorCh <- cmd.Wait()
+	}()
+	return nil
 }
 
 // check validates `h`.  It also creates the disks if needed.
@@ -296,20 +319,6 @@ func (h *HyperKit) check() error {
 		}
 	}
 	return nil
-}
-
-func (h *HyperKit) execute(cmdline string) error {
-	log.Debugf("hyperkit: execute %#v", h)
-	if err := h.check(); err != nil {
-		return err
-	}
-	// Run
-	h.buildArgs(cmdline)
-	cmd, err := h.makeCommand()
-	if err != nil {
-		return err
-	}
-	return h.execHyperKit(cmd)
 }
 
 // Stop the running VM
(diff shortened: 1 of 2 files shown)


akimd:
@rn @ijc @djs55 PTAL.

ijc:
Testify is fine by me. FYI I'm about to go on PTO until next year so I won't be able to have a proper look at this PR until then, sorry. Not sure when the others are back but I'd bet on "next year" for them too.

akimd:
@ijc Enjoy your vacations!

go/hyperkit.go
```diff
@@ -486,8 +485,9 @@ func (h *HyperKit) buildArgs(cmdline string) {
 	log.Debugf("hyperkit: CmdLine: %#v", h.CmdLine)
 }
 
-// Execute hyperkit and plumb stdin/stdout/stderr.
-func (h *HyperKit) execHyperKit() error {
+// makeCommand forges the command to run hyperkit, runs and returns it.
+// It also plumbs stdin/stdout/stderr.
+func (h *HyperKit) makeCommand() (*exec.Cmd, error) {
```
ijc: `makeCommand` to me suggests it constructs the `Command` object but not that it runs it. I think I see why you ant to avoid `exec` and `forkAndExec` is a bit of a mouthful. How about `launchCommand`?

go/hyperkit.go
```diff
@@ -573,11 +573,16 @@ func (h *HyperKit) execHyperKit() error {
...
 		log.Debugf("hyperkit: Cannot write state: %v, killing %v", err, h.Pid)
 		h.process.Kill()
-		return err
+		return nil, err
 	}
+	return cmd, nil
+}
+
+// execHyperKit waits for hyperkit, in background if demanded.
+func (h *HyperKit) execHyperKit(cmd *exec.Cmd) error {
```
ijc: exec is now wrong for this function. How about `waitHyperkit` or even just `wait` (given it is invoked as `h.wait()`)?

go/hyperkit.go
ijc: Or given how it is used how about just `launch` (caller would be `h.launch()`).

go/hyperkit.go
```diff
@@ -131,6 +131,9 @@ type HyperKit struct {
 	// process died unexpectedly.
 	ExtraFiles []*os.File `json:"extra_files"`
 
+	// ErrorCh is where the result of Wait'ing for hyperkit is sent.
+	ErrorCh chan error `json:"-"`
```
ijc: I wonder if this might better fit as something returned by `Start`?

go/hyperkit.go
ijc: I was going patch by patch and the last one resolves this! I should have used the GH review mode instead of comment by comment so I could catch myself...

go/hyperkit.go
ijc: This one is dead and buried by the final commit as well, oops!

ijc:
Overall shape looks good to me, one (non-redundant!) comment/question about where the error channel goes.

go/hyperkit.go
akimd: Hi Ian!

Yes, it also happens to me a lot :)  In that case I something just remove my comment.

go/hyperkit.go
akimd: Yep, I considered this too.  I chose to make it a member to avoid changing the signature (for backward compatibility with the code that already uses Start), but also because it allows to avoid having too many return values.  Finally, at some point I considered being able to provide `Start` with the channel I want it to uses (it could simplify a tiny bit our use in D4M), but really that's hardly a valid argument.

I can do whatever you prefer about this.

ijc: I think, but I'm not sure, that the assignment to `h.ErrorCh` in `FromState` is wrong, since in that case there is nothing to feed the error code into the channel -- you wouldn't call `Start` or `Run` under these circumstances since the process is considered to already be running.

If I'm right then rather than dealing with the question of when exactly `h.ErrorCh` will be non-nil and meaningful I think it would be better to make the channel be part of the return from `Start`. That way only callers of `Start` can have any expectation of having a valid error channel.

It also makes the ownership of the channel clearer, rather than any code which happens to have an `h` in its hand potentially thinking it is allowed to receive from `h.ErrorCh`.

akimd: You are right.

Actually the whole point of FromState is unclear to me: what is it used for?

There's one final advantage that we had for putting the channel in the struct, it's that we are considering running several VMs, in which case it was easier that each vm had its own self-contained structure, rather than having to deal with a slice of structures and a slice of channels.

rn: `FromState` is used to recreate a HyperkIt object from the state directory. So you could start it in the background and then sometime later, in a different process can re-create the object to manage it.

ijc:
(pulling this out from under a resolved/hidden comment thread) 

> Actually the whole point of FromState is unclear to me: what is it used for?

I think the idea was that you could reattach to an existing hyperkit (actually more like a linuxkit state dir I suspect). But I don't see it actually in use in the linuxkit repo so maybe I am wrong, perhaps @rn can remember/clarify.

Or maybe that "feature" is redundant/unused can just be removed?

go/hyperkit.go
ijc: @rn Seems like we commented in parallel! (https://github.com/moby/hyperkit/pull/174#issuecomment-356299316). Is that functionality actually used anywhere?

The wrinkle here is that a object which is "recovered" in this way using `FromState` cannot expose the exit code of the process (since it is not performing the wait on it etc and hyperkit may not even be its child).

ijc:
The questions around the utility of `FromState` etc are unaffected by the changes here, so I've merged.

PR #176, opened 2018-01-10
Go: remove uses of the lock file

akimd:
This is now handled by qcow-tool itself for `compact`.  It was actually not needed for `check`: it is read-only and designed to work on a live disk.

See https://github.com/mirage/ocaml-qcow/pull/106.

### DIFF
--- a/go/disk.go
+++ b/go/disk.go
@@ -35,9 +35,7 @@ type Disk interface {
 	// Ensure creates the disk image if needed, and resizes it if needed.
 	Ensure() error
 	// Stop can be called when hyperkit has quit.  It performs sanity checks, compaction, etc.
-	// It can be passed a lock file which is not-used, but kept alive even if the parent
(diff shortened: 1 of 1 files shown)


akimd:
The failure seems unrelated to my PR.

ijc:
> see the pinata PR ... and docker/pinata#8430.

Most of the people who might read this cannot see this internal repo, please make PR descriptions which are usable/complete with only public information.

akimd:
I removed non public information.

djs55:
This is a follow-on from https://github.com/moby/hyperkit/pull/173 and is safe now that the disk image itself is being locked.

Looks good to me!

ijc:
@akimd please could you also remove the reference to Piñata from the commit message. With that LGTM too.

akimd:
@ijc Done, sorry I missed it.

PR #177, opened 2018-01-10
go: make Disk.getCurrentSize() public

aduermael:
I would like to compare the disk's required size with the current size to eventually remove the disk image before calling `Ensure()` (when desired smaller than current). I just need `getCurrentSize` to be made public.

### DIFF
--- a/go/disk.go
+++ b/go/disk.go
@@ -27,6 +27,8 @@ type Disk interface {
 	SetPath(p string)
 	// GetSize returns the desired disk size.
 	GetSize() int
+	// GetCurrentSize returns the current disk size in MiB.
+	GetCurrentSize() (int, error)
 	// String returns the path.
 	String() string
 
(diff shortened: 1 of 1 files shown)


akimd:
FWIW, dealing with decimal vs. binary scales in d4m made me regret that we hardwired MiB here; so plain bytes would avoid such problems. I don't know if that's something that can be changed in hyperkit.

go/disk.go
```diff
@@ -208,8 +208,8 @@ func (d *RawDisk) create() error {
 	return f.Truncate(int64(d.Size) * mib)
 }
 
-// The current disk size in MiB.
-func (d *RawDisk) getCurrentSize() (int, error) {
+// GetCurrentSize retursn the current disk size in MiB.
```
rn: retursn -> returns

also below

aduermael: Oops, sorry about that, I fixed it. Thank!

aduermael:
thank you @rn!

PR #180, opened 2018-01-17
Go: remove the ExtraFiles field

akimd:
It was introduced for Docker for Mac in d1a5c32a404a86facc5800d962901b5e7a1783d9, so that we could ensure that Docker for Mac's lockfile would be preserved even if it were to crash. This use was removed in 2bedc1969f8c6893678fdef55e90e734c170c42c since djs55 had qcow-tool deal with the lock itself.

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -121,16 +121,6 @@ type HyperKit struct {
 	// connected to. ConsoleStdio and ConsoleFile are supported.
 	Console int `json:"console"`
 
-	// ExtraFiles is exactly exec.Cmd.ExtraFiles.  It specifies
-	// additional open files to be inherited by the hyperkit
-	// process. It does not include standard input, standard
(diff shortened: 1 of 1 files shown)


ijc:
I'm in two minds about this, in principal other downstreams might be using this and it is about as trivial as it is possible to be to maintain. OTOH I'm not aware of any other downstreams who actually do use it (linuxkit doesn't) and AFAICR the Hyperkit itself doesn't actually accept open fds on exec for anything other than stdio (which are handled elsewhere).

I think I'd err on merging (partly since you've gone to the effort of opening it, but also it is trivial to reinstate the functionality if there is another user).

akimd:
I had the same hesitations, but felt somewhat guilty of introducing a feature and then not using it (any longer).

ijc:
I'd forgotten it was you who added it (and comparatively recently too), I think that cements my leaning towards removing...

PR #181, opened 2018-01-17
Go: remove the reloading from json

akimd:
In https://github.com/moby/hyperkit/pull/174#issuecomment-356305949 it was discussed that this feature might be unused and useless.   This PR is meant to sort this out.

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -173,39 +173,6 @@ func New(hyperkit, vpnkitsock, statedir string) (*HyperKit, error) {
 	return &h, nil
 }
 
-// FromState reads a json file from statedir and populates a HyperKit structure.
-func FromState(statedir string) (*HyperKit, error) {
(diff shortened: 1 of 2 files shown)


akimd:
I have no idea whether the use in the sample is a proof that the feature should be maintained, or something that should be removed.

ijc:
@rn once upon a time you planned to use this in linuxkit, but I don't think that ever came to pass. Is it likely to now or shall we drop this stuff?

rn:
drop it for now. Unlikely to use it anytime soon. But the PR failed because not all instances of `FromState`were removed

ijc:
CI says:
```
sample/main.go:7:2: imported and not used: "encoding/json"
```

akimd:
Thanks for the heads up!

rn:
It would be good if you could test this *before* submitting

akimd:
@rn Yes, sorry about this.

PR #184, opened 2018-01-23
CI: also publish the COMMIT

akimd:
So that when we download an artifact we can relate it to a precise commit.


ijc:
I doubled checked that the artifact is correct (it is). Merging

PR #186, opened 2018-01-24
Update circle config to 2.0

ijc:
Somewhat based on https://circleci.com/docs/2.0/migrating-from-1-2/, but needed
some other manual tweaks not mentioned there.

This gives us access to some of the new/more powerful features, but the only things we use it for here is to move the artifacts to the top level avoiding  Needing the click/expand through `Container 0` → `Users/` → `distiller/` → `hyperkit/` → `build/`.


ijc:
Right, finally got this working as desired:
screenshot from 2018-01-25 11-03-19

PR #187, opened 2018-02-16
Fix TRIM/DISCARD of small regions

djs55:
We present a virtual sector size of 512 to the VM but are likely using
a 4096 byte native sector size on the local SSD, therefore we have to
take care to align calls to `fcntl(F_PUNCHHOLE)`.

Previously we would align up the `fp_offset` to the native sector size
and then compute the length of the unaligned initial region we need
to manually zero by `aligned_offset - fp_offset`. Unfortunately this
fails to take into account that the `fp_length` might be smaller than
this, for example:

assume `fp_offset = 1024` and `fp_length = 512` (i.e. TRIM one sector)
=> `aligned_offset = 4096`
=> `len_to_zero = aligned_offset - fp_offset = 3072`

and we will erase `3072` bytes even though the request was for `512`.
We will then go on to calculate a -ve number for the `fcntl(F_PUNCHHOLE)`
call which will return `EINVAL`, but damage has already been done.

This patch caps the size of the initial region to `fp_length`.

### DIFF
--- a/src/lib/block_if.c
+++ b/src/lib/block_if.c
@@ -257,27 +257,40 @@ block_delete(struct blockif_ctxt *bc, off_t offset, off_t len)
 
 		size_t aligned_offset = ALIGNUP(fp_offset, bc->bc_delete_alignment);
 		if (aligned_offset != fp_offset) {
-			size_t len_to_zero = aligned_offset - fp_offset;
-			assert(len_to_zero <= bc->bc_delete_alignment);
-			ssize_t written = pwrite(bc->bc_fd, bc->bc_delete_zero_buf, len_to_zero, (off_t) fp_offset);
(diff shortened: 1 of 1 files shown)


djs55:
I've got the stress test which found this, but it still doesn't run to completion because of an APFS bug. The APFS bug is probably not data-corrupting but it does fail the TRIM which fails the test. Once I've made a workaround for that bug I'll see if I can setup a CI somewhere to run it.

djs55:
I've made some updates following review with @rn. I'm going to try to get the stress test workaround in place now.

FYI @rn I think found a typo in the patch you looked at earlier: guarding the `fcntl` it used to say `if (aligned_length > bc->bc_delete_alignment) {` but I think this should be `if (aligned_length >= bc->bc_delete_alignment) {` (i.e. if the length is exactly equal to 4096, we should still use `fcntl`). I've updated the patch.

PR #191, opened 2018-02-23
Simplify logs

akimd:
- remove useless code
- simplify log redirection
- make sure all the logs are reported as `hyperkit:`

Instead of log redirection for hyperkit(.exe) it would be much more convenient for Docker for Mac that hyperkit used aslog directly, so that its logs are properly credited to it rather than to the caller, which happens to be com.docker.driver.

cc: @djs55

### DIFF
--- a/go/disk.go
+++ b/go/disk.go
@@ -65,7 +65,7 @@ func GetDiskFormat(path string) DiskFormat {
 	case ".raw", ".img":
 		return DiskFormatRaw
 	default:
-		log.Debugf("Unknown disk extension %q, will use raw format", path)
+		log.Debugf("hyperkit: Unknown disk extension %q, will use raw format", path)
 		return DiskFormatRaw
 	}
 }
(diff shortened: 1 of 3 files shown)


akimd:
If agreeable with you guys, I think some parts of the logs could be simplified if we vendored logrus.

ijc:
> If agreeable with you guys, I think some parts of the logs could be simplified if we vendored logrus.

I'm more or less ambivalent about the use of logrus for log messages generated by the go bindings themselves meaning: if it doesn't mess up other users, specifically the `linuxkit` cli, then I'm fine to go ahead. IIRC `linuxkit` CLI uses `logrus` too already, although it has some non-default settings for the non `-v` case which might need a quick check to see if the result remains sane.

I'm a little less sure about the messages from the C hyperkit binary (including various debug message but also any serial devices etc mapped to stdio), but so long as the behaviour when `if h.Console == ConsoleStdio` is retained (and I don't think you are suggesting otherwise) then changing the following `else if log != nil`to follow the same logger as the rest of that code seems sensible (again, assuming the impact on other users, specifically `linuxkit`, are double checked...)

akimd:
@ijc Thanks!

In my comment, I was suggesting to use logrus to replace the stubs that I have already deployed.  It would be less code to simply get rid of the stubs and use directly logrus, which, I believe, is a de facto standard.  In any case I did not mean to change the route of the logs by using logrus.

However, it would definitely make our lives much easier if all of the components we depend upon could be asked to sit on top of Apple's log service.  That's another topic.

Still, I would like this PR to be approved and merged :)  We've been using it in Docker for Mac.

PR #196, opened 2018-03-29
Offer a means to send the tty to the Apple log system

akimd:
On Docker for Windows, all the logs are aggregated: GUI, LinuxKit, etc.  Currently debugging on Docker for Mac is less easy, in particular because LinuxKit's log are sent to the consolefile rings of fixed sized buffers.  As a result, it's hard to find the interleaving of actions between Docker for Mac and LinuxKit, the console files are fragmented and full of '\0', etc.

It is possible to have hyperkit redirect the tty's output to its stderr/stdout, and then its caller (Docker for Mac's driver) could redirect these streams to the logger.  Unfortunately on recent macOS releases, asl is replaced by os_log which uses exclusively the path of the executable as source (changing `argv[0]` is useless).  As a result, hyperkit's logs are "credited" to the driver.

The most elegant approach is to embrace asl/os_log.

Currently Docker for Mac still targets 10.11, and os_log appeared in macOS 10.12, so we cannot use directly os_log's API, we use it through the asl shim.

- add a new device configuration, asl, that sends the tty's output to asl
- in Go, add a new ConsoleLog enum to enum it.

Since some of the logs use escapes for the terminal, escapes appear in the logs.   But that's another issue.

    2018-03-30 10:09:04.784291 +0200	par défaut	75122	com.docker.hyperkit	\^[[34mINFO\^[[0m[0000] starting containerd                           \^[[34mmodule\^[[0m=containerd \^[[34mrevision\^[[0m=9b55aab90508bd389d7654c4baf173a981477d55 \^[[34mversion\^[[0m=v1.0.1
    2018-03-30 10:09:04.785597 +0200	par défaut	75122	com.docker.hyperkit	\^[[34mINFO\^[[0m[0000] loading plugin "io.containerd.content.v1.content"...  \^[[34mmodule\^[[0m=containerd \^[[34mtype\^[[0m=io.containerd.content.v1
    2018-03-30 10:09:04.786756 +0200	par défaut	75122	com.docker.hyperkit	\^[[34mINFO\^[[0m[0000] loading plugin "io.containerd.snapshotter.v1.btrfs"...  \^[[34mmodule\^[[0m=containerd \^[[34mtype\^[[0m=io.containerd.snapshotter.v1
    2018-03-30 10:09:04.788161 +0200	par défaut	75122	com.docker.hyperkit	\^[[33mWARN\^[[0m[0000] failed to load plugin io.containerd.snapshotter.v1.btrfs  \^[[33merror\^[[0m="path /var/lib/containerd/io.containerd.snapshotter.v1.btrfs must be a btrfs filesystem to be used with the btrfs snapshotter" \^[[33mmodule\^[[0m=containerd

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -44,6 +44,8 @@ const (
 	ConsoleStdio = iota
 	// ConsoleFile configures console to a tty and output to a file
 	ConsoleFile
+	// ConsoleLog configures console to a tty and sends its contents to the logs
+	ConsoleLog
 
 	defaultVPNKitSock = "Library/Containers/com.docker.docker/Data/s50"
 
@@ -117,8 +119,7 @@ type HyperKit struct {
 	// Memory is the amount of megabytes of memory for the VM.
 	Memory int `json:"memory"`
 
-	// Console defines where the console of the VM should be
-	// connected to. ConsoleStdio and ConsoleFile are supported.
+	// Console defines where the console of the VM should be connected to.
 	Console int `json:"console"`
 
 	// Below here are internal members, but they are exported so
@@ -199,11 +200,15 @@ func (h *HyperKit) check() error {
 	log.Debugf("hyperkit: check %#v", h)
 	var err error
 	// Sanity checks on configuration
-	if h.Console == ConsoleFile && h.StateDir == "" {
-		return fmt.Errorf("If ConsoleFile is set, StateDir must be specified")
-	}
-	if h.Console == ConsoleStdio && !isTerminal(os.Stdout) && h.StateDir == "" {
-		return fmt.Errorf("If ConsoleStdio is set but stdio is not a terminal, StateDir must be specified")
+	switch h.Console {
+	case ConsoleFile, ConsoleLog:
+		if h.StateDir == "" {
+			return fmt.Errorf("If ConsoleFile or ConsoleLog is set, StateDir must be specified")
+		}
+	case ConsoleStdio:
+		if !isTerminal(os.Stdout) && h.StateDir == "" {
+			return fmt.Errorf("If ConsoleStdio is set but stdio is not a terminal, StateDir must be specified")
+		}
 	}
 	for _, image := range h.ISOImages {
 		if _, err = os.Stat(image); os.IsNotExist(err) {
@@ -420,10 +425,20 @@ func (h *HyperKit) buildArgs(cmdline string) {
 		nextSlot++
 	}
 
-	if h.Console == ConsoleStdio && isTerminal(os.Stdout) {
-		a = append(a, "-l", "com1,stdio")
-	} else {
-		a = append(a, "-l", fmt.Sprintf("com1,autopty=%s/tty,log=%s/console-ring", h.StateDir, h.StateDir))
+	// -l: LPC device configuration.
+	{
+		cfg := "com1"
+		if h.Console == ConsoleStdio && isTerminal(os.Stdout) {
+			cfg += fmt.Sprintf(",stdio")
+		} else {
+			cfg += fmt.Sprintf(",autopty=%s/tty", h.StateDir)
+		}
+		if h.Console == ConsoleLog {
+			cfg += fmt.Sprintf(",asl")
+		} else {
+			cfg += fmt.Sprintf(",log=%s/console-ring", h.StateDir)
+		}
+		a = append(a, "-l", cfg)
 	}
 
 	if h.Bootrom == "" {
@@ -440,15 +455,15 @@ func (h *HyperKit) buildArgs(cmdline string) {
 	log.Debugf("hyperkit: CmdLine: %#v", h.CmdLine)
 }
 
-// openTty opens the tty files for reading, and returns it.
-func (h *HyperKit) openTty() *os.File {
+// openTTY opens the tty files for reading, and returns it.
+func (h *HyperKit) openTTY() *os.File {
 	path := fmt.Sprintf("%s/tty", h.StateDir)
 	for {
 		if res, err := os.OpenFile(path, os.O_RDONLY, 0); err != nil {
-			log.Infof("hyperkit: openTty: %v, retrying", err)
+			log.Infof("hyperkit: openTTY: %v, retrying", err)
 			time.Sleep(10 * time.Millisecond)
 		} else {
-			log.Infof("hyperkit: openTty: got %v", path)
+			log.Infof("hyperkit: openTTY: got %v", path)
 			saneTerminal(res)
 			setRaw(res)
 			return res
@@ -485,7 +500,7 @@ func (h *HyperKit) execute() (*exec.Cmd, error) {
 			cmd.Stderr = os.Stderr
 		} else {
 			go func() {
-				tty := h.openTty()
+				tty := h.openTTY()
 				defer tty.Close()
 				io.Copy(os.Stdout, tty)
 			}()
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -422,7 +422,7 @@ func (h *HyperKit) buildArgs(cmdline string) {
 
 	if h.Console == ConsoleStdio && isTerminal(os.Stdout) {
 		a = append(a, "-l", "com1,stdio")
-	} else if h.StateDir != "" {
+	} else {
 		a = append(a, "-l", fmt.Sprintf("com1,autopty=%s/tty,log=%s/console-ring", h.StateDir, h.StateDir))
 	}
 
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -440,6 +440,22 @@ func (h *HyperKit) buildArgs(cmdline string) {
 	log.Debugf("hyperkit: CmdLine: %#v", h.CmdLine)
 }
 
+// openTty opens the tty files for reading, and returns it.
+func (h *HyperKit) openTty() *os.File {
+	path := fmt.Sprintf("%s/tty", h.StateDir)
+	for {
+		if res, err := os.OpenFile(path, os.O_RDONLY, 0); err != nil {
+			log.Infof("hyperkit: openTty: %v, retrying", err)
+			time.Sleep(10 * time.Millisecond)
+		} else {
+			log.Infof("hyperkit: openTty: got %v", path)
+			saneTerminal(res)
+			setRaw(res)
+			return res
+		}
+	}
+}
+
 // execute forges the command to run hyperkit, runs and returns it.
 // It also plumbs stdin/stdout/stderr.
 func (h *HyperKit) execute() (*exec.Cmd, error) {
@@ -469,20 +485,9 @@ func (h *HyperKit) execute() (*exec.Cmd, error) {
 			cmd.Stderr = os.Stderr
 		} else {
 			go func() {
-				ttyPath := fmt.Sprintf("%s/tty", h.StateDir)
-				var tty *os.File
-				for {
-					var err error
-					tty, err = os.OpenFile(ttyPath, os.O_RDONLY, 0)
-					if err == nil {
-						break
-					}
-					time.Sleep(10 * time.Millisecond)
-				}
-				saneTerminal(tty)
-				setRaw(tty)
+				tty := h.openTty()
+				defer tty.Close()
 				io.Copy(os.Stdout, tty)
-				tty.Close()
 			}()
 		}
 	} else if log != nil {
--- a/src/lib/asl.c
+++ b/src/lib/asl.c
@@ -34,7 +34,7 @@ void asl_put(uint8_t c)
 		asl_log(asl, log_msg, ASL_LEVEL_NOTICE, "%s", buf);
 		buf_idx = 0;
 	} else {
-		if (buf_idx + 2 >= sizeof buf) {
+		if (buf_idx + 2 >= sizeof(buf)) {
 			/* Running out of space, flush.  */
 			asl_put('\n');
 		}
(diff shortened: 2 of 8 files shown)


akimd:
Hi @rn.  Thanks for the tips!

I have reached a point where this is really nice to use from Docker for Mac.

src/lib/asl.c
```diff
@@ -0,0 +1,56 @@
...
+
+/* Grow buf/buf_capacity. */
+static void buf_grow(void)
+{
+	if (buf_capacity == 0) {
+		buf_capacity = 1024;
+		buf = malloc(buf_capacity);
+	} else {
+		buf_capacity *= 2;
+		buf = realloc(buf, buf_capacity);
```
ijc: Both `malloc` and `realloc` can fail so you need to check here. Take extra care with `realloc` not to leak the original buffer -- although given there isn't much error handling you can do without a lot of reworking perhaps an `xhyve_abort` would be a reasonable choice and then leaks are somewhat immaterial.

You could also simplify using the fact that `realloc(NULL, N)` is equivalent to `malloc(N)`.

src/lib/uart_emul.c
```diff
@@ -121,6 +122,7 @@ struct uart_softc {
 
 	struct ttyfd tty;
 	struct log log;
+        bool	asl;		/* Output to Apple logger. */
```
ijc: GH's rendering suggests a tabs vs spaces issue here

src/lib/uart_emul.c
```diff
@@ -773,6 +777,10 @@ uart_set_backend(struct uart_softc *sc, const char *backend, const char *devname
 			if (uart_mapring(sc, logname) == -1) {
 				goto err;
 			}
+		} else if (strcmp("asl", backend) == 0) {
+                        sc->asl = true;
+                        asl_init();
+			retval = 0;
```
ijc: More tab/space issues?

go/hyperkit.go
```diff
@@ -440,6 +455,22 @@ func (h *HyperKit) buildArgs(cmdline string) {
 	log.Debugf("hyperkit: CmdLine: %#v", h.CmdLine)
 }
 
+// openTty opens the tty files for reading, and returns it.
+func (h *HyperKit) openTty() *os.File {
```
ijc: TTY is an abbreviation so either `openTTY` or `opentty` throughout would be preferable.

src/lib/asl.c
akimd: I did not check that allocation succeeded, because there are many places where this is not done, starting with that very file (https://github.com/moby/hyperkit/blob/b2a7a6a1611d3e7489331762c27c7c6d4de73043/src/lib/uart_emul.c#L650).

What would you recommend?  `perror`?

akimd:
I uploaded a version that should address your concerns.  Thanks!

src/lib/asl.c
```diff
@@ -0,0 +1,56 @@
...
+static size_t buf_size = 0;
+static size_t buf_capacity = 0;
+
+/* asl is deprecated in favor of os_log starting with macOS 10.12.  */
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+
+/* Grow buf/buf_capacity. */
+static void buf_grow(void)
+{
+	buf_capacity = buf_capacity ? 2 * buf_capacity : 1024;
```
rn: Why not initialise the buffer capacity to 1024 in `asl_init()` and then just grow from there? Also, we should probably add an upper bound unconditionally flush if it is reached. Or more generally, is there a reason you just don't allocate a fixed buffer of, say 4k and flush it when it's full (or on `\n`). Seem's much simpler.

src/lib/asl.c
```diff
@@ -0,0 +1,56 @@
...
+
+#include <SystemConfiguration/SystemConfiguration.h>
+
+#include <xhyve/asl.h>
+
+static aslclient asl = NULL;
+static aslmsg log_msg = NULL;
+
+static unsigned char *buf = NULL;
+static size_t buf_size = 0;
```
rn: `buf_size` seems misnamed, as it is not the size of the buffer, but rather the index into the buffer where to write.

src/lib/asl.c
```diff
@@ -0,0 +1,56 @@
...
+
+/* Grow buf/buf_capacity. */
+static void buf_grow(void)
+{
+	buf_capacity = buf_capacity ? 2 * buf_capacity : 1024;
+	buf = realloc(buf, buf_capacity);
+	if (!buf) {
+		perror("buf_grow");
+		exit(1);
+	}
```
ijc: I'd have used `xhyve_abort` but this is ok too I guess.

src/lib/uart_emul.c
```diff
@@ -436,6 +438,8 @@ uart_write(struct uart_softc *sc, int offset, uint8_t value)
 			ttywrite(&sc->tty, value);
 			if (sc->log.ring)
 				ringwrite(&sc->log, value);
+			if (sc->asl)
```
rn: below it looks like the `asl` and `log` are exclusive, as a `else if` is used while here it seems as if they can be used together.

src/lib/asl.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+
+/* Send one character to the logger: wait for full lines before actually sending. */
+void asl_put(uint8_t c)
+{
+	if (buf_size + 1 >= buf_capacity) {
+		buf_grow();
+	}
+	if (c == '\n') {
```
ijc: I'm slightly concerned about what happens if the console is e.g. running some sort of curses app or something where `\n` might conceivably be quite rare, it could end up building up quite a large buffer.

Perhaps add an upper bound on the buffer and flush then even if no `\n` is observed?

Also don't you need to handle `\0` specially here? Without such handling the buffer passed to `asl_log` will be truncated at any NUL in the stream. Perhaps just treat `\0` as equivalent to `\n`?

src/lib/asl.c
akimd: That would depend on what you call size.  Size _is_ the size of the contents of the buffer (0 when it's empty).  Capacity is the "available size".

src/lib/asl.c
akimd: There were several `exit(1)` already.  I don't know which coding style to follow, there are too many.

src/lib/asl.c
akimd: I have no idea what can be sent to tty here, whatever you suggest, I'll buy!  I'm fine with your \0 = \n approach, or `\0` = `\\0` to make it visible, etc.

If you are aware of a tty2text approach, I'll take it too.

src/lib/uart_emul.c
akimd: I don't see that they are exclusive.  What do you mean?

src/lib/asl.c
ijc: I think it depends on whether the code was originally kernel, library or application code in the FreeBSD bhyve stack. Looks like `xhyve_abort` is only in "kernel" code so that was a bad suggestion, sorry. In src/lib `abort` and `exit` are basically even so lets stick with `exit`.

src/lib/asl.c
ijc: I think what can be sent to the console here is effectively arbitrary bytes, could be anything, from raw binary to utf-16 encoded text, there's probably no good universal answer. Most frequently it'll be ASCII or UTF-8 though plus some escape codes, I expect.
 
How about:
* Put an upper limit of 4K on the buffer. If we reach this limit then flush unconditionally (could also make it a static 4K buffer if you like since the dynamism is no longer that important with a static limit in place).
* Treat `\n` and `\0` the same: do not inject them in the log stream, instead flush the current buffer.
* For `\` insert it as `\\` (i.e. escape it, to distinguish from the result of the next rule)
* If `!isprint(c)` then escape it as hex `\xx`
* otherwise simply insert `c` as is.

This should preserve all of the useful text in the common case without injecting control characters or anything too weird into ASL (which I don't expect to handle anything especially crazy). It's not quite reversible since we've lost the distinction between `\n` and `\0` but I think it'll do, including the `\n` would make things pretty unreadable in the common case.

The escaping makes the buffer size check a little more complicated, since it's no longer `1:1`. I suggest just growing when you reach `capacity - 3` rather than being too fancy.

rn: If we put an upper limit of 4k on the buffer, why not just statically allocate it?

src/lib/uart_emul.c
rn: sorry, I misread the snippet of code from below. This is fine.

src/lib/asl.c
ijc: Looking at http://www.manpagez.com/man/3/asl_log/ I'm half wondering if perhaps ASL is already setup to deal with such things correctly. There's a bunch of `ASL_ENCODING_*` which look to only be used with `asl_add_output_file` but there seems to be a general implication that the internal DB treats things as NUL terminated strings ("All keys and values are NUL-character terminated C language strings.  UTF-8 encoding may be used for non-ASCII characters.") and it understands escaping in the presentation layer to some extent.

So, while I'm a bit wary that we're going to bring things crashing down around us lets try just flushing on a) 4K limit and b) on `\n` or `\0` and hope for the best.

src/lib/asl.c
```diff
@@ -0,0 +1,44 @@
...
+
+/* Send one character to the logger: wait for full lines before actually sending. */
+void asl_put(uint8_t c)
+{
+	if ((c == '\n') || (c == 0)) {
+		buf[buf_idx] = 0;
+		asl_log(asl, log_msg, ASL_LEVEL_NOTICE, "%s", buf);
+		buf_idx = 0;
+	} else {
+		if (buf_idx + 2 >= sizeof buf) {
```
ijc: Please use `sizeof(buf)`. If it were me I would also prefer to write `buf_idx + 1 > sizeof(buf)` but each to his own.

src/lib/asl.c
```diff
@@ -0,0 +1,44 @@
...
+/* Send one character to the logger: wait for full lines before actually sending. */
+void asl_put(uint8_t c)
+{
+	if ((c == '\n') || (c == 0)) {
+		buf[buf_idx] = 0;
+		asl_log(asl, log_msg, ASL_LEVEL_NOTICE, "%s", buf);
+		buf_idx = 0;
+	} else {
+		if (buf_idx + 2 >= sizeof(buf)) {
+			/* Running out of space, flush.  */
```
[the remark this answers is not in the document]
akimd: Well, `asl_put('\n')` flushes.  When should I start to help us see it?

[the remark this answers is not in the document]
rn: but then you start writing over the end of the buffer...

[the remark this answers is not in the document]
ijc: It took me a moment to grok that too. How about adding `asl_flush()` with the contents of the `\n` case and calling that from both?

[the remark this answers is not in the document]
akimd: @ijc Ok, good idea.  @rn sorry, I don't understand what you mean.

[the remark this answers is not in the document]
ijc: @rn I think `buf_idx + 2 >= sizeof(buf)` ensures there is enough space for the `\0` which would be added for flushing. I'm not quite sure why it is +2 rather than +1 though, with the use of `>=` it seems like it would flush a byte too soon (but, meh).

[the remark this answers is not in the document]
rn: @ijc just pointed out the "subtle" recursive call to `asl_put()`. This is just plain crazy! don't do that.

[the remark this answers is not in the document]
ijc: I suppose intruding on the `asl_*` namespace also used by the OS provided functions is moderately confusing. Perhaps our wrappers should all be `xhyve_asl_*` to avoid that?

src/lib/log.c
```diff
@@ -0,0 +1,50 @@
...
+	log_client = asl_open(NULL, NULL, 0);
+	log_msg = asl_new(ASL_TYPE_MSG);
+}
+
+
+/* Send the content of the buffer to the logger. */
+static void log_flush(void)
+{
+	buf[buf_idx] = 0;
+	asl_log(log_client, log_msg, ASL_LEVEL_NOTICE, "%s", buf);
```
rn: I I read the code which calls this function correctly in neither case where this function is called do you null terminate `buf`. So, how does `asl_log()` know how long `buf` is?

akimd: Hi @rn.  Again, I'm sorry, but I don't know what you mean.  It seems to me that you missed line 32, or I misunderstand your sentence.

rn: sorry, my bad. shouldn't review code without having coffee before

akimd: 😄

src/lib/log.c
```diff
@@ -0,0 +1,50 @@
...
+}
+
+
+/* Send one character to the logger: wait for full lines before actually sending. */
+void log_put(uint8_t c)
+{
+	if ((c == '\n') || (c == 0)) {
+		log_flush();
+	} else {
+		if (buf_idx + 2 >= sizeof(buf)) {
```
rn: this now seems to waste one byte. 

But more in general maybe have something like:
```
if (buf_idx + 2 >= sizeof(buf)) {
    buf[buf_idx++] = c;
    c = 0;
}
if ((c == '\n') || (c == 0)) {
    buf[buf_idx] = 0;
    asl_log(log_client, log_msg, ASL_LEVEL_NOTICE, "%s", buf);
    buf_idx = 0;
} else {
    buf[buf_idx++] = c;
}
```

This would keep everything in one function, but it's a matter of taste, i guess

akimd: Yes, I guess.  I dislike assignments such as `c=0`, and side-effect in expressions such as `buf[buf_idx++]`.  Besides, in some situations you are now adding a `\n` in the buffer.

akimd:
Superseded by #199.

PR #202, opened 2018-04-09
Manually upgrade Python to avoid brew failures

ijc:
It looks like perhaps some sort of upgrade rename around Python/Python 3 has
gone on so I'm hoping this might help.

Deliberately not signed off (yet) because I think our bot is broken and I want to
test that too.


ijc:
I've converted this into a proper workaround (details in the new code comment) for now, please review.

I'll raise a reverting PR as soon as this is merged to provide a canary for the base image being fixed.

PR #203, opened 2018-04-10
ci: remove python/brew workaround.

ijc:
This reverts f55b5442ad65188f0002264a08de38ca47350e4d which was added in #202.

This is not expected to pass CI until circle updates their base image to include the newer brew with default python = 3, I don't know what the typically timeline for that is. In the meantime someone should hit rebuild every few weeks or something.


djs55:
The workaround was deleted in e3718b60c389a3f057a28028e2360180303f2e65 from #265 so I'll close this PR.

Issue #200, opened 2018-04-09
error: format specifies type 'ssize_t' (aka 'long') but the argument has type 'off_t' (aka 'long long')


cmoad:
I am getting the following errors in master. Changing the formatters to `%ll` fixes the problem. Any ideas why I am seeing this? OS 10.13.4

```
☁  hyperkit [master] make -j4
cc src/lib/firmware/bootrom.c
src/lib/firmware/bootrom.c:95:49: error: format specifies type 'ssize_t' (aka 'long') but the argument has type 'off_t' (aka 'long long')
      [-Werror,-Wformat]
                fprintf(stderr, "Invalid bootrom size %zd\n", sbuf.st_size);
                                                      ~~~     ^~~~~~~~~~~~
                                                      %lld
src/lib/firmware/bootrom.c:101:22: error: format specifies type 'ssize_t' (aka 'long') but the argument has type 'off_t' (aka 'long long')
      [-Werror,-Wformat]
                    "page size\n", sbuf.st_size);
                                   ^~~~~~~~~~~~
src/lib/firmware/bootrom.c:112:4: error: format specifies type 'ssize_t' (aka 'long') but the argument has type 'off_t' (aka 'long long')
      [-Werror,-Wformat]
                        sbuf.st_size);
                        ^~~~~~~~~~~~
3 errors generated.
make: *** [build/lib/firmware/bootrom.o] Error 1
```

```
diff --git a/src/lib/firmware/bootrom.c b/src/lib/firmware/bootrom.c
index 02cdecd..3ae2b6a 100644
--- a/src/lib/firmware/bootrom.c
+++ b/src/lib/firmware/bootrom.c
@@ -92,12 +92,12 @@ uint64_t bootrom_load(void)
         * MMIO space (e.g. APIC, HPET, MSI).
         */
        if (sbuf.st_size > MAX_BOOTROM_SIZE || sbuf.st_size < XHYVE_PAGE_SIZE) {
-               fprintf(stderr, "Invalid bootrom size %zd\n", sbuf.st_size);
+               fprintf(stderr, "Invalid bootrom size %lld\n", sbuf.st_size);
                goto done;
        }

        if (sbuf.st_size & XHYVE_PAGE_MASK) {
-               fprintf(stderr, "Bootrom size %zd is not a multiple of the "
+               fprintf(stderr, "Bootrom size %lld is not a multiple of the "
                    "page size\n", sbuf.st_size);
                goto done;
        }
[...]

PR #204, opened 2018-04-10
ci: bump to xcode 9.3 and minimum MacOS to 10.11

ijc:
Also revert the Python workaround from f55b5442ad65188f0002264a08de38ca47350e4d
which was added in #202.

The removal of the Python workaround is optimistic/speculative in case this base image has a newer brew than the old one, if that turns out not to be the case I'll drop that hunk and continue to rely on #203.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,11 +2,11 @@ version: 2
 jobs:
   build:
     macos:
-      xcode: "9.0"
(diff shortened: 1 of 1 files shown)


ijc:
CI hits #200/#201, will rebase once that lands.

ijc:
Python issue isn't fixed by this newer version, so I dropped that hunk.

PR #208, opened 2018-04-16
go: Don't check for initrd

rn:
hyperkit happily boots with just a kernel (and the rootfs on
a disk device, for example). So, if booting in non-fw mode
only check for the kernel.


ijc:
Should we not check that iff `h.Initrd != ""` it exists?

rn:
@ijc, good idea. I pushed an update

PR #212, opened 2018-05-04
Send (stdout/stderr) logs to the system logger

akimd:
Currently in Docker for Mac, the various backend tools use the system logger, which is extremely convenient to use, both from the CLI or from its GUI (Console.app).  However, some of the log messages from hyperkit are still just sent to stdout/stderr, and they then appear credited to the parent process (com.docker.driver).  For instance:

```
2018-05-11 11:30:31.930181 +0200	par défaut	60743	com.docker.hyperkit	[  287.446631] reboot: Power down
2018-05-11 11:30:31.933951 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: virtio-net-vpnkit: initialising, opts="path=vpnkit.eth.sock,uuid=d0460c75-ebae-4c39-bc51-83914bbc93b6"
2018-05-11 11:30:31.934027 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: virtio-net-vpnkit: interface will have uuid d0460c75-ebae-4c39-bc51-83914bbc93b6
2018-05-11 11:30:31.934062 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: virtio-net-vpnkit: Connection established with MAC=02:50:00:00:00:01 and MTU 1500
2018-05-11 11:30:31.934118 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: virtio-9p: initialising path=vpnkit.port.sock,tag=port
2018-05-11 11:30:31.934178 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: linkname vms/0/tty
2018-05-11 11:30:31.934241 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: COM1 connected to /dev/ttys004
2018-05-11 11:30:31.934287 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: stdout: COM1 linked to vms/0/tty
2018-05-11 11:30:32.008685 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: failed to read stderr: EOF
2018-05-11 11:30:32.008809 +0200	par défaut	60740	com.docker.driver.amd64-linux	hyperkit: failed to read stdout: EOF
2018-05-11 11:30:33.374676 +0200	par défaut	60843	com.docker.driver.amd64-linux	hyperkit: stderr: virtio-net-vpnkit: magic=VMN3T version=22 commit=0123456789012345678901234567890123456789\^B
2018-05-11 11:30:33.375147 +0200	par défaut	60843	com.docker.driver.amd64-linux	hyperkit: stderr: vsock init 3:0 = vms/0, guest_cid = 00000003
2018-05-11 11:30:36.358223 +0200	par défaut	60846	com.docker.hyperkit	\^[[2J\^[[01;01H\^[[=3h\^[[2J\^[[01;01H\^[[2J\^[[01;01H\^[[=3h\^[[2J\^[[01;01H\^[[2J\^[[01;01H\^[[=3h\^[[2J\^[[01;01H\^[[2J\^[[01;01H\^[[=3h\^[[2J\^[[01;01H\^[[2J\^[[01;01H\^[[=3h\^[[2J\^[[01;01H\^[[0m\^[[39m\^[[49m\^[[2J\^[[01;01H\^[[=3h\^[[2J\^[[01;01H\^[[0m\^[[39m\^[[49m\^[[2J\^[[01;01H\^[[0m\^[[39m\^[[49mWelcome to GRUB!
2018-05-11 11:30:36.358543 +0200	par défaut	60846	com.docker.hyperkit	
2018-05-11 11:30:36.366057 +0200	par défaut	60846	com.docker.hyperkit	
\^[[0m\^[[39m\^[[49m\^[[0m\^[[39m\^[[49m\^[[2J\^[[01;01H\^[[0m\^[[39m\^[[49m  Booting `LinuxKit ISO Image'
2018-05-11 11:30:36.366112 +0200	par défaut	60846	com.docker.hyperkit	
2018-05-11 11:30:36.588090 +0200	par défaut	60846	com.docker.hyperkit	
[    0.000000] Linux version 4.9.87-linuxkit-aufs (root@70f48253c8b4) (gcc version 6.4.0 (Alpine 6.4.0) ) #1 SMP Fri Mar 16 18:16:33 UTC 2018
2018-05-11 11:30:36.589129 +0200	par défaut	60846	com.docker.hyperkit	[    0.000000] Command line: BOOT_IMAGE=/boot/kernel console=ttyS0 page_poison=1 vsyscall=emulate panic=1 root=/dev/sr0 text
2018-05-11 11:30:36.589856 +0200	par défaut	60846	com.docker.hyperkit	[    0.000000] x86/fpu: Supporting XSAVE feature 0x001: 'x87 floating point registers'
2018-05-11 11:30:36.590459 +0200	par défaut	60846	com.docker.hyperkit	[    0.000000] x86/fpu: Supporting XSAVE feature 0x002: 'SSE registers'
...
2018-05-11 11:30:37.737567 +0200	par défaut	60846	com.docker.hyperkit	[    1.104987] NET: Registered protocol family 1
2018-05-11 11:30:37.738268 +0200	par défaut	60846	com.docker.hyperkit	[    1.105549] pci 0000:00:1f.0: Activating ISA DMA hang workarounds
2018-05-11 11:30:37.738530 +0200	par défaut	60843	com.docker.driver.amd64-linux	hyperkit: stderr: 
rdmsr to register 0x64e on vcpu 0
2018-05-11 11:30:37.738592 +0200	par défaut	60843	com.docker.driver.amd64-linux	hyperkit: stderr: rdmsr to register 0x34 on vcpu 0
2018-05-11 11:30:37.739582 +0200	par défaut	60846	com.docker.hyperkit	[    1.106691] RAPL PMU: API unit is 2^-32 Joules, 5 fixed counters, 163840 ms ovfl timer
```
All these logs are now credited to hyperkit (without the clutter of the `hyperkit: stdout:` prefixes).

To this end, these commits:
- enforce some consistency in the use of stdout/stderr
- offer a CLI option to decide whether we stick to stderr, or use the system logger
- offer a Go option to this CLI option

This PR has a peer in Docker for Mac: https://github.com/docker/pinata/pull/9309.

### DIFF
--- a/src/hyperkit.c
+++ b/src/hyperkit.c
@@ -145,6 +145,7 @@ usage(int code)
 		    "       -h: help\n"
 		    "       -H: vmexit from the guest on hlt\n"
 		    "       -l: LPC device configuration. Ex: -l com1,stdio -l com2,autopty -l com2,/dev/myownpty\n"
+		    "       -L: destination for logs: 'stderr' (default), or 'log'\n"
 		    "       -m: memory size in MB, may be suffixed with one of K, M, G or T\n"
 		    "       -P: vmexit from the guest on pause\n"
 		    "       -s: <slot,driver,configinfo> PCI slot config\n"
@@ -869,7 +870,7 @@ main(int argc, char *argv[])
 	rtc_localtime = 1;
 	fw = 0;
 
-	while ((c = getopt(argc, argv, "behvuwxMACHPWY:f:F:g:c:s:m:l:U:")) != -1) {
+	while ((c = getopt(argc, argv, "behvuwxMACHPWY:f:F:g:c:s:m:l:L:U:")) != -1) {
 		switch (c) {
 		case 'A':
 			acpi = 1;
@@ -901,6 +902,9 @@ main(int argc, char *argv[])
 				errx(EX_USAGE, "invalid lpc device "
 				    "configuration '%s'", optarg);
 			}
+			break;
+                case 'L':
+			log_set_destination(optarg);
 			break;
 		case 's':
 			if (pci_parse_slot(optarg) != 0)
--- a/src/hyperkit.c
+++ b/src/hyperkit.c
@@ -132,42 +132,45 @@ static uint64_t (*fw_func)(void);
 __attribute__ ((noreturn)) static void
 usage(int code)
 {
-
-        fprintf(stderr,
-                "Usage: %s [-behuwxMACHPWY] [-c vcpus] [-F <pidfile>] [-g <gdb port>] [-l <lpc>]\n"
-		"       %*s [-m mem] [-p vcpu:hostcpu] [-s <pci>] [-U uuid] -f <fw>\n"
-		"       -A: create ACPI tables\n"
-		"       -c: # cpus (default 1)\n"
-		"       -C: include guest memory in core file\n"
-		"       -e: exit on unhandled I/O access\n"
-		"       -f: firmware\n"
-		"       -F: pidfile\n"
-		"       -g: gdb port\n"
-		"       -h: help\n"
-		"       -H: vmexit from the guest on hlt\n"
-		"       -l: LPC device configuration. Ex: -l com1,stdio -l com2,autopty -l com2,/dev/myownpty\n"
-		"       -m: memory size in MB, may be suffixed with one of K, M, G or T\n"
-		"       -P: vmexit from the guest on pause\n"
-		"       -s: <slot,driver,configinfo> PCI slot config\n"
-		"       -u: RTC keeps UTC time\n"
-		"       -U: uuid\n"
-		"       -v: show build version\n"
-		"       -w: ignore unimplemented MSRs\n"
-		"       -W: force virtio to use single-vector MSI\n"
-		"       -x: local apic is in x2APIC mode\n"
-		"       -Y: disable MPtable generation\n",
-		progname, (int)strlen(progname), "");
-
+	if (code) {
+		printf("Usage: %s [-behuwxMACHPWY] [-c vcpus] [-F <pidfile>] [-g <gdb port>] [-l <lpc>]\n"
+		    "       %*s [-m mem] [-p vcpu:hostcpu] [-s <pci>] [-U uuid] -f <fw>\n"
+		    "       -A: create ACPI tables\n"
+		    "       -c: # cpus (default 1)\n"
+		    "       -C: include guest memory in core file\n"
+		    "       -e: exit on unhandled I/O access\n"
+		    "       -f: firmware\n"
+		    "       -F: pidfile\n"
+		    "       -g: gdb port\n"
+		    "       -h: help\n"
+		    "       -H: vmexit from the guest on hlt\n"
+		    "       -l: LPC device configuration. Ex: -l com1,stdio -l com2,autopty -l com2,/dev/myownpty\n"
+		    "       -m: memory size in MB, may be suffixed with one of K, M, G or T\n"
+		    "       -P: vmexit from the guest on pause\n"
+		    "       -s: <slot,driver,configinfo> PCI slot config\n"
+		    "       -u: RTC keeps UTC time\n"
+		    "       -U: uuid\n"
+		    "       -v: show build version\n"
+		    "       -w: ignore unimplemented MSRs\n"
+		    "       -W: force virtio to use single-vector MSI\n"
+		    "       -x: local apic is in x2APIC mode\n"
+		    "       -Y: disable MPtable generation\n",
+		    progname, (int)strlen(
+			    progname), "");
+	} else {
+		fprintf(stderr,
+		    "Try '%s --help' for more information.\n", progname);
+	}
 	exit(code);
 }
 
 __attribute__ ((noreturn)) static void
 show_version()
 {
-        fprintf(stderr, "%s: %s\n\n%s\n",progname, VERSION,
+        printf("%s: %s\n\n%s\n", progname, VERSION,
 		"Homepage: https://github.com/docker/hyperkit\n"
 		"License: BSD\n");
-		exit(0);
+	exit(0);
 }
 
 void
--- a/src/lib/log.c
+++ b/src/lib/log.c
@@ -1,24 +1,28 @@
 #include <asl.h>
-#include <pwd.h>
 #include <fcntl.h>
+#include <pwd.h>
 #include <stdio.h>
+#include <sysexits.h>
 #include <time.h>
 
 #include <SystemConfiguration/SystemConfiguration.h>
 
 #include <xhyve/log.h>
 
+/* Some functions below invoke functions with non-literal format
+ * strings, and it's ok: their (top-level) fmt string was checked. */
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+
 static aslclient log_client = NULL;
 static aslmsg log_msg = NULL;
 
-static unsigned char buf[4096];
+static char buf[4096];
 /* Index of the _next_ character to insert in the buffer. */
 static size_t buf_idx = 0;
 
 /* asl is deprecated in favor of os_log starting with macOS 10.12.  */
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 
-/* Initialize ASL logger and local buffer. */
 void log_init(void)
 {
 	log_client = asl_open(NULL, NULL, 0);
@@ -36,7 +40,7 @@ static void log_flush(void)
 
 
 /* Send one character to the logger: wait for full lines before actually sending. */
-void log_put(uint8_t c)
+static void log_put(char c)
 {
 	if ((c == '\n') || (c == 0)) {
 		log_flush();
@@ -46,5 +50,98 @@ void log_put(uint8_t c)
 		}
 		buf[buf_idx] = c;
 		++buf_idx;
+	}
+}
+
+
+/* Send a string to the logger: wait for full lines before actually sending. */
+static void log_puts(const char *s)
+{
+	for (; *s; ++s) {
+		log_put(*s);
+	}
+}
+
+
+/* Send a string to the logger: wait for full lines before actually sending. */
+static int log_vprintf(const char *fmt, va_list args)
+{
+	static char buf[4096];
+
+	int res = vsnprintf(buf, sizeof(buf), fmt, args);
+
+	log_puts(buf);
+	return (res);
+}
+
+
+/* Destination for logs.  */
+static enum {
+	LOG_DST_LOG,
+	LOG_DST_STDERR
+}
+log_dst = LOG_DST_STDERR;
+
+void log_set_destination(const char *dst)
+{
+	if (!dst) {
+		log_fprintf(stderr,
+		    "log_set_destination: invalid NULL argument");
+		exit(EX_USAGE);
+	} else if (!strcmp(dst, "log")) {
+		log_dst = LOG_DST_LOG;
+	} else if (!strcmp(dst, "stderr")) {
+		log_dst = LOG_DST_STDERR;
+	} else {
+		log_fprintf(stderr, "log_set_destination: invalid argument: %s",
+		    dst);
+		exit(EX_USAGE);
+	}
+}
+
+
+int log_fprintf(FILE *f, const char *fmt, ...)
+{
+	va_list args;
+	int res = 0;
+
+	va_start(args, fmt);
+	if ((log_dst == LOG_DST_LOG) &&
+	    ((f == stdout) || (f == stderr))) {
+		res = log_vprintf(fmt, args);
+	} else {
+		res = vfprintf(f, fmt, args);
+	}
+	va_end(args);
+	return (res);
+}
+
+
+/* A buffer for the console.  */
+static char console_buf[4096];
+/* Index of the _next_ character to insert in console_buf. */
+static size_t console_idx = 0;
+
+
+/* Send the content of the buffer to the logger. */
+static void log_console_flush(void)
+{
+	console_buf[console_idx] = 0;
+	log_fprintf(stderr, "%s\n", console_buf);
+	console_idx = 0;
+}
+
+
+/* Send one character to the logger: wait for full lines before actually sending. */
+void log_console_put(char c)
+{
+	if ((c == '\n') || (c == 0)) {
+		log_console_flush();
+	} else {
+		if (console_idx + 2 >= sizeof(console_buf)) {
+			log_console_flush();
+		}
+		console_buf[console_idx] = c;
+		++console_idx;
 	}
 }
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -121,6 +121,10 @@ type HyperKit struct {
 
 	// Console defines where the console of the VM should be connected to.
 	Console int `json:"console"`
+	// LogDestination specifies where the logs go: `stderr` or
+	// `log` (denoting the platform's log system).  Empty denotes
+	// `stderr`.
+	LogDestination string `json:"log_destination"`
 
 	// Below here are internal members, but they are exported so
 	// that they are written to the state json file, if configured.
@@ -209,6 +213,11 @@ func (h *HyperKit) check() error {
 		if !isTerminal(os.Stdout) && h.StateDir == "" {
 			return fmt.Errorf("If ConsoleStdio is set but stdio is not a terminal, StateDir must be specified")
 		}
+	}
+	switch h.LogDestination {
+	case "", "stderr", "log":
+	default:
+		return fmt.Errorf("invalid LogDestination: %v", h.LogDestination)
 	}
 	for _, image := range h.ISOImages {
 		if _, err = os.Stat(image); os.IsNotExist(err) {
@@ -441,6 +450,10 @@ func (h *HyperKit) buildArgs(cmdline string) {
 			cfg += fmt.Sprintf(",log=%s/console-ring", h.StateDir)
 		}
 		a = append(a, "-l", cfg)
+	}
+
+	if h.LogDestination == "log" {
+		a = append(a, "-L", "log")
 	}
 
 	if h.Bootrom == "" {
--- a/src/include/xhyve/log.h
+++ b/src/include/xhyve/log.h
@@ -1,7 +1,19 @@
 #pragma once
 
-/* Initialize ASL logger and local buffer. */
+/* Initialize logger. */
 void log_init(void);
 
-/* Send one character to the logger: wait for full lines before actually sending. */
-void log_put(uint8_t _c);
+/* Send one character to the console logger. */
+void log_console_put(char _c);
+
+/* Specify where logs should be sent: `stderr` or `log`.  Dies on error.  */
+void log_set_destination(const char* dst);
+
+/* If logging is enabled, intercept outputs to stdout/stderr to the
+ * logger, otherwise forward to fprintf. */
+__attribute__ ((format (printf, 2, 3)))
+int log_fprintf(FILE* f, const char *fmt, ...);
+
+/* Intercept all the calls to fprintf to honor log_set_destination.  */
+#define fprintf log_fprintf
+#define printf(...) log_fprintf(stdout, __VA_ARGS__)
--- a/src/include/xhyve/xhyve.h
+++ b/src/include/xhyve/xhyve.h
@@ -78,3 +78,5 @@ void vcpu_add(int fromcpu, int newcpu, uint64_t rip);
 int fbsdrun_vmexit_on_hlt(void);
 int fbsdrun_vmexit_on_pause(void);
 int fbsdrun_virtio_msix(void);
+
+#include <xhyve/log.h>
--- a/src/lib/uart_emul.c
+++ b/src/lib/uart_emul.c
@@ -439,7 +439,7 @@ uart_write(struct uart_softc *sc, int offset, uint8_t value)
 			if (sc->log.ring)
 				ringwrite(&sc->log, value);
 			if (sc->asl)
-				log_put(value);
+				log_console_put((char)value);
 		} /* else drop on floor */
 		sc->thre_int_pending = true;
 		break;


rn:
Please don’t post WIP PRs for cosmetic changes like this. For changes like this it should be possible to submit a proper PR, with descriptions and appropriate subject. Anyone watching this repository gets notifications, CI gets kicked off etc. As a maintainer I have no idea what to do with PRs like this. Do you want feedback? Do you want me to review it know, or is it going to change?

akimd:
> Please don’t post WIP PRs for cosmetic changes like this.

I'm sorry that the fact that I work, and how I work, bothers you.  And I do appreciate your evaluation of the nature of my work.  Always a pleasure.

> For changes like this it should be possible to submit a proper PR, with descriptions and appropriate subject.

It turns out that it's not so easy to work on a D4D component alone without checking that it works properly in D4D.  I'm sure you know that.  Besides, there are always surprises with CI.  I'm sure you also know that.  For instance, here Circle CI fails, and I have no idea why.

> Anyone watching this repository gets notifications, CI gets kicked off etc.

The CI is the whole point, both for checks, but also to feed that hyperkit into D4D's own CI.  If notifications are a trouble for you, maybe you can disable them, or ignore them, like we all do for notifications we are not directly concerned by.

> As a maintainer I have no idea what to do with PRs like this. Do you want feedback? Do you want me to review it know, or is it going to change?

"Work in process" means "don't bother checking now, people are working here".  When ready, it will get all the things a proper PR requires, be sure about that.  But before that, well, I gotta work.

rn:
For simple changes like this you really can just do the work locally and then submit a PR. You can even run the test locally and replace the Docker for Mac binary on your system if you want to test there. The hyperkit CI does not run any Docker for Mac tests and all the tests it executes are in the Makefile.

Tests in CI may still fail, but at least you can test everything locally before submitting a PR.

For larger changes it is perfectly fine to submit an RFC style patch (or open an issue to discuss it) to get feedback, but for relatively simple changes like this, in the future, please only submit PRs when the code is ready for review.

thebsdbox:
@akimd  The circleCI is failing as `#include <xhyve/log.h>` is missing from `src/lib/mirage_block_c.c` so the strict compiler settings complain about `log_info()` not being declared.

rn:
@thebsdbox previously (on Friday) it was failing on something else:
```
hdiutil: mount failed - no mountable file systems
    while executing
"exec hdiutil mount -mountpoint ./test-vol ${DISK}s1"
    (file "./test_linux.exp" line 28)
```
which is what I had seen before (and it magically and annoyingly went away when enabling `ssh` access).

But yes, now it fails on a compile error

akimd:
Back to the original error.
```
Requesting system halt

sd 0:0:0:0: [sda] Synchronizing SCSI cache

sd 0:0:0:0: [sda] Stopping disk

reboot: System haltedhdiutil: mount failed - no mountable file systems
    while executing
"exec hdiutil mount -mountpoint ./test-vol ${DISK}s1"
    (file "./test_linux.exp" line 28)

make: *** [test] Error 1
Exited with code 2
```

rn:
Downloaded the DMG and ran `hdiutil imageinfo` on it:
```
Backing Store Information:
	URL: file:///Users/rneugeba/Downloads/disk.dmg
	Name: disk.dmg
	Class Name: CBSDBackingStore
Class Name: CRawDiskImage
Checksum Type: none
Size Information:
	Total Bytes: 20971520
	Compressed Ratio: 1
	Sector Count: 40960
	Total Non-Empty Bytes: 20971520
	Compressed Bytes: 20971520
	Total Empty Bytes: 0
Format: UDRW
Format Description: raw read/write
Checksum Value: 
Properties:
	Encrypted: false
	Kernel Compatible: true
	Checksummed: false
	Software License Agreement: false
	Partitioned: false
	Compressed: no
Segments:
	0: /Users/rneugeba/Downloads/disk.dmg
partitions:
	partition-scheme: fdisk
	block-size: 512
	partitions:
		0:
			partition-name: Master Boot Record
			partition-start: 0
			partition-synthesized: true
			partition-length: 1
			partition-hint: MBR
			boot-code: 0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
		1:
			partition-start: 1
			partition-number: 1
			partition-length: 40959
			partition-hint: DOS_FAT_32
			partition-filesystems:
				FAT16: NO NAME
	burnable: false
Resize limits (per hdiutil resize -limits):
40960	40960	1123981568
```
looks reasonably sane to me

rn:
Also performed on the DMG:
```
% hdiutil attach -nomount -noverify -noautofsck disk.dmg | head -n1 | cut -f1 -d " "
/dev/disk2
% hdiutil mount -mountpoint ./test-vol /dev/disk2s1
% ls -l test-vol/
total 0
-rwxrwxrwx  1 rneugeba  staff  0  9 May 11:46 test*
```
which is basically what the `expect` script does

ijc:
We wait for the vm to output "reboot: System halted" but I wonder if we could be racing between the hyperkit process exiting (and flushing any remaining IO) and the `hdiutil mount`?

I think adding a `wait` (no args) right before the `exec hdiutil mount` would cause it to wait for the process spawned by the previous `spawn` (which is hyperkit). @akimd it's a bit of a long shot but could you add that here please.

Is there a Macos "sysinfo" we could log the output of to get things like the version of the OS?

ijc:
I notice that we mount the disk.dmg, which produces `/dev/disk«n»` _and_ `/dev/rdisk«n»` (former is a block special device, latter is an unbuffered thing or some sort). We pass the `rdisk` variant to hyperkit, but the `hdiutil mount` uses the `disk` variant, so I wonder if there is some cache inconsistency we are tripping over? I can't see any `sync` subcommand of `hdiutil`, so maybe we need to detach/reattach (this is an even longer shot than above though)

akimd:
There is this:
```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.13.4
BuildVersion:   17E202
```
Where would you want it?

ijc:
> Where would you want it?

Just running that command a a step in the circle config so it appears in the CI logs would be fine, thanks.

I'm theorising that maybe newer macos has differing behaviour which is showing up a prexisting race, or something

ijc:
The `wait` just blocked until circle killed it. Suggesting that either the current child process (from expect's point of view) isn't Hyperkit (i.e. I was wrong to suggest using it) or hyperkit isn't actually exiting.

ijc:
Running locally with the `wait` added does succeed, suggesting the modification to the script is correct but for some reason hyperkit is not exiting in CI (or is exiting _very_ slowly).

ijc:
Running the script under `dtruss` (reminding me how much I hate both `dtruss` and "system integrity protection") shows it is `wait`ing on the same process `fork`ed. So I think the modification to add the wait is correct.

rn:
I created https://github.com/moby/hyperkit/pull/214 to better test this in isolation. I added some logging, but missed some of the convo here

src/lib/log.c
```diff
@@ -48,3 +49,68 @@ void log_put(uint8_t c)
...
 	}
 }
+
+
+// Destination for logs.
+static enum {
+	log_dst_log,
+	log_dst_stderr
+}
+log_dst = log_dst_stderr;
```
rn: This really confused me because the `log_dst = log_dst_stderr;` was on a new line. Please pull this onto the same line as the `}`. Also, according to `CODINGSTYLE.md` please use all upper case for the values.

src/lib/pci_ahci.c
```diff
@@ -105,16 +105,14 @@ enum sata_fis_type {
  * Debug printf
  */
 #ifdef AHCI_DEBUG
-static FILE *dbg;
 #define DPRINTF(format, ...) \
 	do { \
-		fprintf(dbg, format, __VA_ARGS__); \
-		fflush(dbg); \
+		log_info(format, __VA_ARGS__); \
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/pci_virtio_9p.c
```diff
@@ -21,7 +21,7 @@
 #define VIRTIO_9P_MOUNT_TAG 1
 
 static int pci_vt9p_debug = 0;
-#define DPRINTF(params) if (pci_vt9p_debug) printf params
+#define DPRINTF(params) if (pci_vt9p_debug) log_info params
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/pci_virtio_block.c
```diff
@@ -115,7 +115,7 @@ struct virtio_blk_hdr {
  * Debug printf
  */
 static int pci_vtblk_debug;
-#define DPRINTF(params) if (pci_vtblk_debug) printf params
+#define DPRINTF(params) if (pci_vtblk_debug) log_info params
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/pci_virtio_net_vmnet.c
```diff
@@ -139,7 +139,7 @@ struct virtio_net_rxhdr {
  * Debug printf
  */
 static int pci_vtnet_debug;
-#define DPRINTF(params) if (pci_vtnet_debug) printf params
+#define DPRINTF(params) if (pci_vtnet_debug) log_info params
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/pci_virtio_sock.c
```diff
@@ -124,9 +124,9 @@ struct virtio_sock_hdr {
  * Debug printf
  */
 static int pci_vtsock_debug = 0;
-#define DPRINTF(params) do { if (pci_vtsock_debug) { printf params; fflush(stdout); } } while(0)
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/pci_virtio_net_tap.c
```diff
@@ -119,8 +119,8 @@ struct virtio_net_rxhdr {
  * Debug printf
  */
 static int pci_vtnet_debug;
-#define DPRINTF(params) if (pci_vtnet_debug) printf params
-#define WPRINTF(params) printf params
+#define DPRINTF(params) if (pci_vtnet_debug) log_info params
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/pci_virtio_rnd.c
```diff
@@ -53,8 +53,8 @@
 
 
 static int pci_vtrnd_debug;
-#define DPRINTF(params) if (pci_vtrnd_debug) printf params
```
rn: there shouldn't be any need to change this debug code. this won't be enabled for release builds and there should be no need to redirect to a logger.

src/lib/log.c
```diff
@@ -48,3 +49,68 @@ void log_put(uint8_t c)
...
+
+void log_set_destination(const char *dst)
+{
+	if (!dst) {
+		log_error("log_set_destination: include NULL argument");
+		exit(EX_USAGE);
+	} else if (!strcmp(optarg, "log")) {
```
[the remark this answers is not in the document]
akimd: Yep!

src/lib/log.c
akimd: I agree, but the code was indented by uncrustify.  I don't fight tools.

src/lib/pci_ahci.c
akimd: I thought it would help the developers to have a consistent behavior.  I'll remove that.

src/hyperkit.c
```diff
@@ -133,8 +133,7 @@ __attribute__ ((noreturn)) static void
 usage(int code)
 {
 
-        fprintf(stderr,
-                "Usage: %s [-behuwxMACHPWY] [-c vcpus] [-F <pidfile>] [-g <gdb port>] [-l <lpc>]\n"
+        log_info("Usage: %s [-behuwxMACHPWY] [-c vcpus] [-F <pidfile>] [-g <gdb port>] [-l <lpc>]\n"
```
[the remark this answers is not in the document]
akimd: That's exactly what I did.  I'll upload that soon, I'm on the D4D side currently.

src/lib/log.c
rn: tools are there to aid not to dictate. please move it to the line with the `}`. The current form just looks wrong an confusing and style(9) as examples with `typedef structs` which have the more natural way as well.

src/lib/pci_ahci.c
akimd: Well, it turns out that there is still a lot of "debug" messages (DPRINTF) that do show on D4D as of today.  They should be part of hyperkit's log like the others.  So I will make again a proposal for these debug logs.

rn:
A more general comment/question. What does this PR try to achieve which can't be done within the wrapper, we already have? Can't the wrapper just redirect stdout/stderr to ASL? My general concern is that this PR changes a lot of files which are from upstream bhyve. Changing code in the middle of these files, which this PR does, may make cherry-picking back-porting changes unnecessarily harder.

akimd:
I understand this concern.  I had considered redefining `fprintf` to control this.  But being able to tell apart mere logs from actual errors is really an added value.

How about the converse: shouldn't we submit these changes upstream?

rn:
I still don't understand why this can't be done in the wrapper. HyperKit is hardly chatty and only really prints warnings/errors

akimd:
I don't know what you call the wrapper, but as I have already reported it about the console output, the Apple system logger watches the name of the process (and does not even let itself be fooled by games on argv[0]).

ijc:
What's so bad with things which are logged as `com.docker.wrapper: hyperkit: «some message from hyperkit»` (where `com.docker.wrapper` is whatever D4M process spawns hyperkit and babysits its i/o).

akimd:
@ijc Are you really asking the question?  How often did you have to read logs of Docker for Mac?  I regularly consumes these logs, and I can tell you that anything that can help to make the logs be more readable is a time savior.

ijc:
I have worked extensively on Docker 4 Mac in the past, and in particular on the integration of Hyperkit there. I've spent as much time as anyone going through docker for mac support logs and certainly in the case of specifically diagnosing Hyperkit issues in that context probably far more than the average.

In any case, Rolf has, as a Hyperkit maintainer, expressed a dislike for the amount of divergence of the code from the upstream bhyve code which is happening here which will make future cherry picking and syncing harder. With that same hat on I agree with his assessment.

This PR is not going to be merged while it introduces such levels of divergence with bhyve. Unless you can reduce this to one or two isolated changes then I'm afraid I think this PR has no future and you will need to figure out a way to arrange logging to your liking externally to hyperkit. If you don't want to use the existing supervisor process then perhaps via a new intermediary process dedicated to handling hyperkit logs for the purposes of labelling them or whatever solution you prefer.

akimd:
Yes, I definitely understand the worries about conflicts, but the cherry-picks I had seen were not that extensive.

akimd:
I have checked on Docker for Mac's side, and intercepting `fprintf` works well.  I should probably also intercept `printf`, but at least at this point we get nice logs.

WDYT?

src/lib/log.c
```diff
@@ -31,15 +31,57 @@ void log_init(void)
...
+		log_info("%s", buf+4);
+	} else if (has_prefix(buf, "WARN[")) {
+		log_warn("%s", buf+4);
+	} else if (has_prefix(buf, "ERRO[")) {
+		log_error("%s", buf+4);
+	} else if (strcasestr(buf, "cannot") ||
+	    strcasestr(buf, "could not") ||
+	    strcasestr(buf, "error") ||
+	    strcasestr(buf, "failed") ||
+	    strcasestr(buf, "unable to")) {
```
ijc: Isn't `buf` the guest console log? How can you know enough about its structure and contents (in general) to assert that hardcoding all these heuristics makes sense at the Hyperkit level?

src/lib/log.c
```diff
@@ -31,15 +31,57 @@ void log_init(void)
 }
 
 
+/* Remove tty escapes for colors from buf.  */
+static void remove_escapes()
```
ijc: The lack of any bounds checking in this function is a bit of a concern. If this is safe by construction due to other places in the code then at least a comment to explain how/why is needed. But even better would be to arrange that this function is locally safe and not capable of running off the end of the array.

src/lib/log.c
```diff
@@ -151,6 +193,16 @@ void log_info(const char *fmt, ...)
...
 
 
+void log_warn(const char *fmt, ...)
+{
+	va_list args;
+
+	va_start(args, fmt);
+	log_vlog(ASL_LEVEL_WARNING, fmt, args);
+	va_end(args);
+}
```
ijc: this seems to have strayed into a different patch to the addition of the other `log_{debug,info,err}` functions.

src/lib/log.c
ijc: My `Isn't buf the guest console log?` further down applies to the massaging done here too I think.

src/lib/log.c
```diff
@@ -48,3 +103,111 @@ void log_put(uint8_t c)
...
+
+
+/* Send a string to the logger: wait for full lines before actually sending. */
+static int log_vprintf(const char *fmt, va_list args)
+{
+	static char buf[4096];
+
+	int res = vsnprintf(buf, sizeof(buf), fmt, args);
+
+	log_puts(buf);
```
ijc: Do I understand correctly that the log buffer will now be a mixture of the guest console output and any diagnostics emitted by Hyperkit itself using `fprintf`? I am very much not keen on that arrangement.

Logs from Hyperkit itself should either (depending on the configuration) go to stdio or to ASL directly, never via the log buffering stuff used for the guest console.

It looks like the direct uses of `log_info` and friends are doing the right thing here, it's just the imposition on the `fprintf(stderr, ...` calls which looks wrong.

ijc:
I should also reiterate that I have yet to see anything here which couldn't also be achieved by hyperkit's parent process nor any convincing arguments why it should not be done that way.

src/lib/log.c
akimd: Hi Ian!
There's no guarantee, it's purely a heuristics.  Log level are purely "decorators", they don't trigger actions as Panicf would do in go.  Log levels help drawing the attention to some events, and also decide whether the event is volatile or persistent.
So I'd rather have maybe _some_ false positive, than flat logs with a single level.
Given that we don't want massive changes in hyperkit, that's a reasonable heuristics.

src/lib/log.c
akimd: I'm not sure I follow here: the treatment only strips contents, so it cannot go wrong.  Do you want to me to add an `assert(dst <= src)`?

src/lib/log.c
akimd: I've moved it there, thanks.

src/lib/log.c
akimd: I've made it the other way round: the console lines are assembled until the end of line, and then we send it to the logger.  And the logger sends them to stderr or asl, depending on the command line choice.

akimd:
> I should also reiterate that I have yet to see anything here which couldn't also be achieved by hyperkit's parent process nor any convincing arguments why it should not be done that way.

Well, it's easy to have convincing arguments that you can do something — just do it — but the opposite is somewhat harder...

So, asl is phased out, replaced by oslog since 10.12.  The asl API is still there, working, but I venture it's a shim to oslog.  And oslog doesn't care about the `sender` and all these entries that allowed to "lie" about the source of the logs.  The only thing that appears in the logs is the name of the process (the real one, playing with argv[0] is useless).  That's why in order to have hyperkit's log/console appear in the logger under the hyperkit banner, we have to do from hyperkit.

src/lib/log.c
ijc: I'm afraid we cannot accept these sorts of application specific heuristics into a low level and generic component such as Hyperkit, so this will have to go.

src/lib/log.c
ijc: I'm concerned that you might run off the end of `buf` not that `dst` might get ahead of `src`. Nothing in this function ensures you can't keep going over the end of the buf AFAICT.

ijc: Actually, this is all moot since as with the strcmp heuristic to pick a log level this sort of manipulation of the console stream has no place in Hyperkit so there is no point debating the implementation.

ijc:
> That's why in order to have hyperkit's log/console appear in the logger under the hyperkit banner, we have to do from hyperkit.

No, that is simply an arbitrary requirement you have set for yourself. The hyperkit process logs (as opposed to the console logs, which it is useful to have separately and which already go to ASL) could appear to come from a parent process just fine, there isn't so many of them that they will make the logs unusable.

So, there are some bits here which are acceptable (the core idea of passing the `fprintf(stderr` into the ASL logs, without otherwise touching them, is tollerable even with the necessary #define based hack) but the application specific manipulation of the log stream and the heuristics around log level are not acceptable for hyperkit and this PR cannot be merged with them.

akimd:
Hi!
As demanded, I have removed (all) the logs massaging.  As a consequence, the log_info etc. are useless, and removed too. There remains only the redirection of fprintf to the logger.
What do you think about that?

akimd:
What can I do to go this go forward?

ebriney:
@ijc : Can we forward on that? The PR just adds a flag to redirect logs and make them clean in console.app and in our d4m diagnostics, so I don't really see why don't merge it.
What must we change to get it merged and avoid to fork/vendor the repo to get this in d4m?

rn:
We have stated repeatedly what is needed:
- https://github.com/moby/hyperkit/pull/212#pullrequestreview-120262616
- https://github.com/moby/hyperkit/pull/212#issuecomment-389186728
- https://github.com/moby/hyperkit/pull/212#issuecomment-388871860
- https://github.com/moby/hyperkit/pull/212#issuecomment-388356415
- https://github.com/moby/hyperkit/pull/212#issuecomment-388332640
- https://github.com/moby/hyperkit/pull/212#issuecomment-388331090
- ...

This has not sufficiently been addressed and I will ignore this PR until there is a concise description of the issue this PR is trying to solve and why this can't be fixed in the supervisor process, especially since HyperKit is not very verbose at all.

ebriney:
A [convincing argument](https://github.com/moby/hyperkit/pull/212#pullrequestreview-120262616) is that it's more integrated to mac than it was previously
The other comments you mention had been addressed from what I see, no?
If there are new modifications we need to address let's tell us or close the the PR and we will fork the repo, the ci and so on...

PR #214, opened 2018-05-09
CI TEST: DO NOT MERGE

rn:


test/test_linux.exp
```diff
@@ -4,8 +4,26 @@ set KERNEL "./vmlinuz"
 set INITRD "./initrd.gz"
 set CMDLINE "earlyprintk=serial console=ttyS0"
 
+# Get hdiutil info
+set INFO [exec hdituil info -debug]
```
ijc: You've misspelled `hdiutil` throughout.

rn:
Ta, happens when doing this while in a meeting and cut and pasting...

ijc:
I've been running `test_linux.exp` with an added `wait` under `dtruss` (which involved `cp /usr/bin/expect .` then `sudo dtruss ./expect ./test_linux.exp` to workaround integrity protection).

Out of about 40 runs I seem to have caught a hang (perhaps the same one as seen here) twice, so a very low rate compared with here and #212. The output is: 
[log.txt](https://github.com/moby/hyperkit/files/1987997/log.txt)
. I'm not following forks but i can see:
```
fork()		 = 38832 0
...
wait4(0x97B0, 0x7FFEE8695628, 0x0)		 = 38832 0
```

Which I think is the `spawn` and `wait` respectively. Interestingly the `wait4` seems to have completed.

The hyperkit process looks to be a zombie (but maybe that's because of `dtruss` running):
```
nug:test ijc$ pstree  -p 37661
-+= 00001 root /sbin/launchd
 \-+= 35701 root sshd: ijc [priv] 
   \-+- 35703 ijc sshd: ijc@ttys001,ttys003 
     \-+= 37661 ijc -bash
       |-+= 38826 root sudo dtruss ./expect ./test_linux.exp
       | \-+- 38828 root /bin/sh /usr/bin/dtruss ./expect ./test_linux.exp
       |   \-+- 38830 root /usr/sbin/dtrace -x dynvarsize 4m -x evaltime preinit -Z -n \012 #pragma D option quiet\012 \012 
       |     \-+- 38831 root ./expect ./test_linux.exp
       |       \--= 38848 root (hyperkit)
       \--- 38827 ijc tee log
```
and
```
$ ps aux | grep [3]8848
root            38848   0.0  0.0        0      0   ??  ?Es   2:33pm   0:00.00 (hyperkit)
```
the log ends:
```
workq_kernreturn(0x4, 0x0, 0x0)          = 0 Err#-2
select(0x7, 0x70000377BE30, 0x70000377BD30, 0x70000377BDB0, 0x0)                 = 1 0
kevent_qos(0xFFFFFFFFFFFFFFFF, 0x7FFEE8694740, 0x1)              = 0 0
workq_kernreturn(0x40, 0x7000036E6B80, 0x1)              = 0 Err#-2
select(0x7, 0x70000377BE30, 0x70000377BD30, 0x70000377BDB0, 0x0)                 = 1 0
psynch_mutexdrop(0x7FE1D2C07120, 0x371B03, 0x371A00)             = 0 0
psynch_mutexwait(0x7FE1D2C07120, 0x371B03, 0x371900)             = 3611395 0
kevent_qos(0xFFFFFFFFFFFFFFFF, 0x7FFEE8694770, 0x1)              = 0 0
workq_kernreturn(0x40, 0x7000036E6B80, 0x1)              = 0 Err#-2
workq_kernreturn(0x20, 0x0, 0x1)                 = 0 0
workq_kernreturn(0x4, 0x0, 0x0)          = 0 Err#-2
workq_kernreturn(0x4, 0x0, 0x0)          = 0 0
__disable_threadsignal(0x1, 0x0, 0x0)            = 0 0
dtrace: error on enabled probe ID 2356 (ID 897: syscall::thread_selfid:return): invalid user access in action #5 at DIF offset 0
```
and that final line appeared some time after the hang (unsure if it was because I did "something").

rn:
I think CircleCI runs in VMs so timing will be different there

ijc:
log.txt suggests that tcl is just sitting in an event loop calling select, unclear what it is waiting for.

It's possible my wait is buggy -- the `expect` which looks for `reboot: halted` might have seen the EOF and done an implicit close. I'm trying with this instead:
```diff
diff --git a/test/test_linux.exp b/test/test_linux.exp
index 67b626d..58da9b8 100755
--- a/test/test_linux.exp
+++ b/test/test_linux.exp
@@ -25,6 +25,11 @@ expect {
   "reboot: System halted"
 }
 
+expect {
+  timeout {puts "FAIL process-exit"; exec kill -9 $pid; exit 1}
+  eof
+}
+
 exec hdiutil mount -mountpoint ./test-vol ${DISK}s1
 exec cat ./test-vol/test
 exec hdiutil detach $DISK
```

ijc:
With the above `expect { .... eof }` I don't seem to have a hang any more, so I think the use of `wait` was wrong/racy. I've not seen any disk image corruption as seen in CI though.

I loaded up my system with a handful of `while : ; do : ; done` loops to simulate impared performance and still didn't repro.

akimd:
Great news!

rn:
Hmm, adding a few `hdiutil info` bits, like this PR does, seems to get past the point where it normally fails. Some racey stuff. I'll try @ijc change next, after removing the debug

ijc:
I've done 100 iterations on my local machine with `expect { ... eof }` and no hang or corruption. I'll try without and see what happens.

rn:
After convo with @ijc, no force-pushed an unmodified exp script. Theory is that it might be related to the CI node we are running on...

rn:
So much for that theory (unless we got onto a new VM after the force push), the test fails with no mods in the expect script. Now adding `expect { ... eof }` bit

ijc:
> So much for that theory (unless we got onto a new VM after the force push),

...or the extra `hdiutil`  invocations perturbed stuff enough to work.

ijc:
I went to apply the fix to `test_linux_qcow.exp` too and noticed that it has a `close` at exactly this point.

I think the right thing to do is replace that close by the `expect {... eof }` construct (which has an implicit `close` when it sees the `eof`).

Having reread the docs for `close` and `wait` as well as the stuff on `eof` handling I think we might want a `wait` after the `{expect ... eof}` too. I'm just testing that locally before I raise a PR.

PR #215, opened 2018-05-09
tests: wait for hyperkit to exit before examining the disk image

ijc:
We were seeing cases in CI where the subsequent examination of the disk was
failing with `hdiutil: mount failed - no mountable file systems`.

It was hypothesised (and tested in #212 and #214) that perhaps Hyperkit was not
completely shutting down (and flushing any outstanding I/O) before the script
was examining the disk.

Add an `expect` call to both tests which waits for `eof` from the fds attached
to the Hyperkit process (`eof` here means the other end has shutdown). In the
qcow case this replaces the existing `close` since `expect` does a close in the
`eof` case already.

Once Hyperkit has exited we can `wait` to reap the process, this might be
unnecessary since the test script isn't far from exiting itself which will end
up collecting all orphaned children but we might as well and it gives some
extra confidence that Hyperkit really has gone away.

To help diagnose any future problems we also collect both disk images as
artifacts (they are reasonably small).


akimd:
It has not been merged.  Nor rebuilt.

ijc:
I'm going to try taking a copy of the image before running `hdiutil` and see if that captures the corruption. Then I think I'll try a `sync` and if that doesn't work I'll try reattaching the image between the VM shutdown and the check.

ijc:
The `cp` seems to avoid the corruption. I'm going to try a `sync` next.

ijc:
While adding the host level `sync` I noticed that the `qcow2` test did a guest level `sync` before shutdown while the dmg based one did not -- I so I thought I would try adding that first.

ijc:
Worth mentioning I don't really expect the guest based `sync` to do much, if neither `umount` nor `shutdown` involve doing as sync at some point I'd be shocked.

ijc:
Unexpectedly the guest level `sync` seems to have worked. I'm going to hit rebuild a bunch of times to try and confirm.

ijc:
... and indeed now it fails, good! Will try the host level `sync` next.

ijc:
With "test: sync after vm has exited." added it has passed 5 times in a row (with a 6th ongoing). I hope this means things are ok. @rn maybe you could pull this PR and push to your one just to get confirmation on a possibly different CI node?

rn:
@ijc pushed your changes to my PR and CI is running. Will hit rebuild a couple of time on it when done for good measure

akimd:
Is there any lesson learned here on the shutdown process that should be ported into Docker for Mac?

ijc:
I think (and this is just a theory) the issue is due to block/page cache inconsistencies between the two backend devices produced by `hciutil attach` combined, perhaps, with performance implications from running hyperkit on top of virtualised macos.

That said -- there would be no (or very little) harm in you doing a `sync` from your supervisor whenever hyperkit is shutdown.

PR #217, opened 2018-06-01
go: strip the possible options before checking if the image exists

akimd:
If passed "foo.iso,ro", we fail with "foo.iso,ro does not exist".

But maybe we should always set ",ro" for ISOs?

### DIFF
--- a/go/hyperkit.go
+++ b/go/hyperkit.go
@@ -211,6 +211,10 @@ func (h *HyperKit) check() error {
 		}
 	}
 	for _, image := range h.ISOImages {
+		// Strip possible options.
+		if end := strings.LastIndexByte(image, ','); end != -1 {
+			image = image[0:end]
+		}
 		if _, err = os.Stat(image); os.IsNotExist(err) {
 			return fmt.Errorf("ISO %s does not exist", image)
 		}


justincormack:
There is no point specifying `ro` for an ISO there is no way to write to them.

akimd:
I agree, but then it would be not to `LOCK_EX` in this case.

https://github.com/moby/hyperkit/blob/05c926bcd7466887ca2545757c8c8029e0d9ce7d/src/lib/block_if.c#L715

What fix would you suggest?

akimd:
Another option would be to have `pci_ahci_init` pass `,ro` itself to `blockif_open`.

https://github.com/moby/hyperkit/blob/05c926bcd7466887ca2545757c8c8029e0d9ce7d/src/lib/pci_ahci.c#L2331

Does that sound better to you?  Currently in hyperkit it doesn't not matter that it's an ISO, it's opened with RW perms, and an exclusive lock.  Which prevents concurrent readonly accesses.

akimd:
Closing in favor of https://github.com/moby/hyperkit/pull/219, which is merged.  The CI has not finished yet, but it should pass.

PR #218, opened 2018-06-05
Don't lock RO block device backing files and open ISOs as RO

rn:
This PR contains two small related changes.

The first commit makes sure that HyperKit does not take an advisory exclusive lock on a block device's backing file if `,ro` is passed as an option. The exclusive lock was added with https://github.com/moby/hyperkit/pull/173 to prevent disk corruption when more than one HyperKit process is trying to use the same backing file. For read-only backing files this is not necessary.

The second change adds `,ro` to the options if a `ahci-cd` device is specified. CD devices (ie ISO files) are read-only by definition so we might as well open them read-only (and don't take an exclusive lock).

Together these two changes enable multiple HyperKit instances to share the same ISO file.

I'm not entirely sure if I like adding `,ro` to the `ahci-cd` options. We could simply ask the user to specify `-s 3,ahci-cd,<path to iso>,ro` for ISO images, but that seems tedious as ISOs are read-only. Defaulting `ahci-cd` devices to read-only in different ways would require more invasive changes, such as changing `blockif_open()`.

### DIFF
--- a/src/lib/block_if.c
+++ b/src/lib/block_if.c
@@ -712,9 +712,11 @@ blockif_open(const char *optstr, const char *ident)
 		}
 		/* Lock the file to prevent concurrent write+write or read+write as this
 		   would usually lead to corruption */
-		if (flock(fd, LOCK_NB | (ro ? LOCK_SH : LOCK_EX)) < 0) {
-			perror("Could not lock backing file");
-			goto err;
+		if (!ro) {
+			if (flock(fd, LOCK_NB | (ro ? LOCK_SH : LOCK_EX)) < 0) {
+				perror("Could not lock backing file");
+				goto err;
+			}
 		}
 		if (fstat(fd, &sbuf) < 0) {
 			perror("Could not stat backing file");
--- a/src/lib/pci_ahci.c
+++ b/src/lib/pci_ahci.c
@@ -2402,7 +2402,17 @@ pci_ahci_hd_init(struct pci_devinst *pi, char *opts)
 static int
 pci_ahci_atapi_init(struct pci_devinst *pi, char *opts)
 {
-	return (pci_ahci_init(pi, opts, 1));
+	char nopts[MAXPATHLEN + 3];
+	int ret;
+
+	/* atapi is only used for CDROMs. Add 'ro' to the options. */
+	ret = snprintf(nopts, sizeof(nopts), "%s,ro", opts);
+	if (ret < 0) {
+		WPRINTF("Path to ISO too long (%d)\n", ret);
+		return (ret);
+	}
+	
+	return (pci_ahci_init(pi, nopts, 1));
 }
 
 /*


src/lib/block_if.c
```diff
@@ -712,9 +712,11 @@ blockif_open(const char *optstr, const char *ident)
 		}
 		/* Lock the file to prevent concurrent write+write or read+write as this
 		   would usually lead to corruption */
-		if (flock(fd, LOCK_NB | (ro ? LOCK_SH : LOCK_EX)) < 0) {
-			perror("Could not lock backing file");
-			goto err;
+		if (!ro) {
+			if (flock(fd, LOCK_NB | (ro ? LOCK_SH : LOCK_EX)) < 0) {
```
ijc: There is a now redundant `ro ?` within this check. But I'm not sure about this change. It seems to me that a shared lock is the correct thing for a ro file since while multiple readers are ok you want to exclude any writers from coexisting with those readers. Having `ro` use `LOCK_SH` and writers use `LOCK_EX` (i.e. the status quo) ought to achieve this I think

It seems to me that the existing code was trying to do what we want &mdash; if it is broken it'd be nice to know why, in any case I don't think removing all locking from the `ro` case is likely to be what we really want.

ijc:
Some atapi devices are writeable, but we don't support CDRW ATAPI extensions (I don't think) so that's pretty much moot.

I don't think adding a `ro` argument to `blockif_open` (essentially just move the existing local variable to a parameter and leave the option parsing the same) will be invasive (there is only two callers) and is a cleaner solution IMHO.

Have `bhyve` dealt with this at all? If so we should strongly consider following them.

src/lib/block_if.c
akimd: I agree, the previous code does work.  All it needs is `ro` to be set.

src/lib/block_if.c
```diff
@@ -712,9 +712,11 @@ blockif_open(const char *optstr, const char *ident)
...
+				perror("Could not lock backing file");
```
akimd: For the records, it is annoying that all these messages don't even name the file on which they fail.

rn:
The HyperKit code (as well as the `bhyve` code) only select ATAPI for CD drives, and as you say, we don't support CD writers, ATAPI in our case is synonymous RO.  

`bhyve` does not lock devices, this is HyperKit specific, and ISOs are opened RW unless the user specifies `ro`.

I had started with adding `ro` to `blockif_open()` but then went down the other route. Happy to go back to adding the flag. This won't be too disruptive a change compared to current upstream, where `pci_ahci.c` as diverged a little to support multi-port `ahci` adapters (in addition to the `ahci-hd` and `ahci-cd` options HyperKit already supports). With the new `ahci` option you can specify a mix of `hd` and `cd` devices on the same controller (see https://github.com/freebsd/freebsd/blob/master/usr.sbin/bhyve/pci_ahci.c#L2330). Changing `blockif_open()` to take a `ro` flag shouldn't make it too cumbersome port these changes if we wanted to.

src/lib/pci_ahci.c
```diff
@@ -2402,7 +2402,17 @@ pci_ahci_hd_init(struct pci_devinst *pi, char *opts)
 static int
 pci_ahci_atapi_init(struct pci_devinst *pi, char *opts)
 {
-	return (pci_ahci_init(pi, opts, 1));
+	char nopts[MAXPATHLEN + 3];
```
akimd: I don't know what guarantees MAXPATHLEN gives on the `\0` terminator.  `PATH_MAX` seems to have replaced it. And in hyperkit, it is the latter which is used.

rn:
closing this and will do a new PR whih add `ro` to `blockif_open()`

PR #223, opened 2018-06-21
Log the error if hv_vm_create fails

djs55:
hv_vm_create may fail with

- HV_BAD_ARGUMENT
- HV_BUSY
- HV_NO_DEVICE
- HV_NO_RESOURCES
- HV_UNSUPPORTED

While we already handled `HV_NO_DEVICE` carefully, for the rest we would abort with `hv_vm_create failed` which isn't very helpful. This patch adds a `printf` of the error code so we can investigate a little more (see for example [docker/for-mac#3015])


djs55:
Decoding the other known errors is a good idea. I'll force push a change.

Issue #227, opened 2018-08-22
Allow the use of vmnet as an unprivileged user


stevenjm:
Hi there!

I have a patch that I've been using on my own laptop for a few days to allow the use of `vmnet` as an unprivileged user. It works by opening the `vmnet` interface early and then dropping privileges if it detects that it is running setuid. `chmod u+s` on the `hyperkit` binary is then sufficient to allow unprivileged use of `vmnet`.

There are obviously security concerns with making the `hyperkit` binary setuid. As far as I can tell, the worst thing a user should be able to do is create a VM with the same UUID that another user wants to use so that a `vmnet` interface is created with the same MAC address, blocking the creation of the other user's VM. I'm not sure if this is worth worrying about or not.

I'll be opening a pull request with my proposed implementation of this momentarily.

PR #228, opened 2018-08-22
Allow the use of vmnet as an unprivileged user

stevenjm:
This is my implementation of https://github.com/moby/hyperkit/issues/227.


ijc:
I've not reviewed but I am rather windy (to say the least) about in any way countenancing installing `hyperkit` as `setuid` to root, it's not been even close to audited for that use case and I would bet it is leakier than a rusty sieve...

stevenjm:
There are already various places around the web where it is suggested to make `hyperkit` setuid in order to get `vmnet` to work. Even if it isn't made setuid by default, this ensures that it drops privileges as soon as possible if someone uses that approach.

Perhaps a safer solution would be a standalone setuid binary that does nothing except set up a `vmnet` interface and then `sendfd` it back to `hyperkit`, as that would be easier to audit. I'd like to have some indication that the general idea would be accepted by the project before I do any more work on this, though.

ijc:
I think there's quite a difference between random "advice" on the Internet and the hyperkit project telling users to do something. I also fear that the privilege dropping will do nothing more than give people a false sense of security.

A small standalone binary might well be better, although I don't immediately see in `vmn_create` an fd which could be passed, unless `xpc_object_t` or `interface_ref` are guaranteed to be suitably fd like. Unfortunately the docs link in the comment is now dead :-(

PR #229, opened 2018-08-29
Go: build and return HyperKit CmdLine without requiring Start

ekcasey:
This is useful for developers who want to use the hyperkit Go wrapper to build a hyperkit command but don't want the wrapper to execute that command.

Signed-off-by: Emily Casey <ecasey@pivotal.io>


go/hyperkit.go
```diff
@@ -353,6 +353,12 @@ func (h *HyperKit) String() string {
 	return string(s)
 }
 
+// Command builds and returns a single string of the command line
+func (h *HyperKit) Command(cmdline string) string {
+	h.buildArgs(cmdline)
+	return h.CmdLine
```
ijc: I'm not sure that `h.CmdLine` is actually usable for what you need -- it has collapsed the individual elements of `argv` into a single string (without any quoting) and it's not entirely clear that you can always reverse that..

PR #235, opened 2018-11-24
opam: use --disable-sandboxing

djs55:
In the CircleCI environment we see failures like
```
+ /Users/distiller/.opam/opam-init/hooks/sandbox.sh "build" "mkdir" "-p" "/Users/distiller/.opam/4.03.0/lib/ocaml/" (CWD=/Users/distiller/.opam/4.03.0/.opam-switch/build/ocaml-base-compiler.4.03.0)
- mkdir: /Users/distiller/.opam/4.03.0/lib/ocaml/: Operation not permitted
[ERROR] The compilation of ocaml-base-compiler failed at
        "/Users/distiller/.opam/opam-init/hooks/sandbox.sh build mkdir -p
        /Users/distiller/.opam/4.03.0/lib/ocaml/".
```
This patch attempts to work around these by disabling the new sandboxing
features of `opam`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,7 @@ jobs:
       - run: brew upgrade python
       - run: brew install opam
       - run: opam init --disable-sandboxing -v -n --comp="${OPAM_COMP}" --switch="${OPAM_COMP}" local "${OPAM_REPO}"
+      - run: opam install depext -y
       - run: opam config exec -- opam depext -i hyperkit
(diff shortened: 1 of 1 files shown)


djs55:
I think I need to upgrade the `opam` package metadata to version 2.0 format.

djs55:
At some point we should probably upgrade from opam 1 to opam 2 (or use another solution like a vendoring tool). For now I think we can close this one.

PR #244, opened 2019-01-29
README update: opam install qcow.0.10.4

chrishiestand:
trying to install qcow.0.10.3 yields a dependency problem:

> The following dependencies couldn't be met:
>   - qcow-tool → qcow >= 0.10.4
>   - qcow-tool → sha = 1.9 → ocaml < 4.06.0
>       base of this switch (use `--unlock-base' to force)
> Your request can't be satisfied:
>   - No available version of qcow satisfies the constraints
> 
> No solution found, exiting

But qcow.0.10.4 works

PR #245, opened 2019-01-30
opam: update to qcow.0.10.4 (and sha.1.10)

djs55:
Newer OCaml versions require a newer `qcow` version, see #244.

This PR:
- uses the `1.2` branch of the upstream `opam-repository`. When we update to `opam` 2 we can use the `master` branch again
- bumps `qcow` to `0.10.4` and `sha` to `1.10`


repo/update-upstream.sh
```diff
@@ -22,8 +22,9 @@ cleanup() {
 ## Fetch the opam-repository
 
 cd "${WORK_DIR}"
-git clone --depth=1 ${OPAM_REPO} ${REPO_DIR_NAME}
+git clone ${OPAM_REPO} ${REPO_DIR_NAME}
 cd ${REPO_DIR_NAME}
+git checkout 1.2
```
ijc: FWIW I think this change is equivalent to 
```
go clone --depth=1 --branch=1.2 ${OPAM_REPO} ${REPO_DIR_NAME}
```
which retains the shallow clone. Not sure if `1.2` here is a branch or a tag, but `--branch` is documented to accept either.

ijc: Bah!

I suppose I could edit my previous comment to make you look like a loon but I won't ;-)

djs55:
I think this is good to merge, and then #244 on top.

PR #249, opened 2019-04-05
circleci: use a patched opam.rb

ijc:
The current one gives the following error:

    Error: opam: "cxx11" is not a recognized standard

I cargo culted this change from https://github.com/moby/vpnkit/commit/4bb313b0d76fccc508299133e36c2c015184eec8


ijc:
I agree, but yes, it is what it is.

djs55:
Yes sorry about the current `opam` 1.2 hack. I need to update everything to version 2.

Issue #253, opened 2019-06-28
vpnkitname of 'auto' references legacy default vpnkit socket


yinzara:
The "defaultVPNKitSock" of "$HOME/Library/Containers/com.docker.docker/Data/s50" no longer coincides with the location used on the current version of Docker for Mac.

It has now changed to "$HOME/Library/Containers/com.docker.docker/Data/vpnkit.eth.sock"

This makes setting the vpnkit socket name to "auto" will not work with the current version of Docker for Mac.

yinzara:
See https://github.com/linuxkit/linuxkit/commit/415d0d3c32d43618033e45cc8e2a43a76f7c6001 for another project affected by the same

PR #254, opened 2019-06-28
Fixed auto location of vpnkit socket

yinzara:
Fixes https://github.com/moby/hyperkit/issues/253


yinzara:
This PR maintains legacy behavior if the older versions of Docker for Mac are installed but supports the new location if the file exists.

PR #259, opened 2019-08-20
Workaround problems with hardened runtime, needed for notarization

djs55:
On macOS 10.15 Catalina and later applications need to be [Notarized](https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution#3087722) where notarization is an Apple malware- and security-check of the released binaries.

As part of the notarization process, Apple require that [hardened runtime](https://help.apple.com/xcode/mac/current/#/devf87a2ac8f) is enabled. This prevents
- code injection into running processes
- attaching debuggers
- allocating writable + executable memory

Currently `hyperkit` uses `valloc` to allocate memory for the VM and then grants the VM `READ`, `WRITE` and `EXECUTE`, which will fail if the hardened runtime is enabled.

One workaround is to grant the `hyperkit` binary the [Allow Unsigned Executable Memory Entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-unsigned-executable-memory) which disables the writable+executable check for all allocations done by the process.

This patch proposes another workaround, which is to switch from using `valloc` to `mmap` with the special flag `MAP_JIT`. This allows us to use the weaker [Allow Execution of JIT-compiled Code](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit) entitlement, so that only the VM memory allocation is writable+executable, other allocations are not.

Note that, according to the mono project https://github.com/mono/mono/commit/a502768b3a24f4251de6a48ba78a27c898968e63 the `MAP_JIT` flag causes problems with older version of macOS, so they recommend only enabling it for Mojave and later.

Note that enabling the hardened runtime and adding entitlements is done at the `codesign` stage which means we can't easily test this from the current CI as the binaries are unsigned.


src/lib/vmm/vmm_mem.c
```diff
@@ -29,14 +29,55 @@
...
+		xhyve_abort("vmm_get_kern_osrelease failed to parse sysctl kern.osrelease value\n");
+	}
+	return v;
+}
 
 int
 vmm_mem_init(void)
 {
+	if (vmm_get_kern_osrelease() >= OSRELEASE_MOJAVE) {
+		fprintf(stderr, "Detected Mojave or later, will use MAP_JIT\n");
```
justincormack: I would log on the pre Mojave case, as thats the "buggy" case

justincormack:
This seems reasonable to me. Have you tested that it works with this?

Saviq:
FWIW alternatively one can use the following entitlement:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>com.apple.security.cs.disable-executable-page-protection</key>
    <true/>
</dict>
</plist>
```

This is working for us in Multipass for a while now.

PR #263, opened 2019-10-24
Explicitly disable VMCS shadowing

agustingianni:
On macOS Catalina VM creation fails with:

vmx_set_ctlreg: cap_field: 2 bit: 14 unspecified don't care
vmx_init: processor does not support desired secondary processor-based controls
Unable to create VM (22)

Add the VMCS shadow control bit to the list of disabled controls.

Original patch by John Coyle https://github.com/machyve/xhyve/commit/0c3c50a971f4fdc0d25cbeb06cdecce8fbaeef92


justincormack:
Thanks. Curious about your setup as we haven't had any reports of failures with Catalina, and we have tests running for it, so your environment must differ in some way (different hardware?).

justincormack:
Ah, my machine of that date suffered from the exploding battery issue and its still with Apple, but from the other thread looks like there is a feature missing in some older CPUs.

We are taking a look at the build errors which are unrelated.

agustingianni:
Let me know if you need me to run some tests. Cheers.

justincormack:
Can you rebase, we fixed the CI issue.

agustingianni:
I just rebased with upstream. Hope I did not mess it up. Thanks!

PR #264, opened 2019-10-24
OCaml: upgrade from opam 1 to opam 2

djs55:
`opam` version 1 has been dead for a while and we had been relying on an old `brew` bottle to install it. This bottle depended on `python` version 2 which is also about to disappear.

This PR updates to `opam` version 2 and removes the old "mini opam repository" which contained a snapshot of dependency package metadata. Instead it seems simpler to have a "lock file" `lock.opam` which has concrete versions.

It may be better to use this tool: https://github.com/rgrinberg/opam-lock but I've not investigated it yet.


djs55:
Without the `brew` update (which is currently broken) it installs `opam` version 1 anyway! Maybe a simpler fix will do for now.

PR #266, opened 2019-11-13
Fix issue 260

amaumene:
Fix implicit conversion changes signedness and higher order bits are zeroes after implicit conversion issues. Most probably due to upgrade to Catalina.

PR #261, opened 2019-10-13
Fix implicit conversion changes signedness errors with clang 11

jarrodconnolly:
Fixes #260 using example provided in comments from xhyve project.

Builds and local `make test` passes, I am not an expert in this code so needs some review.

PR #270, opened 2020-02-16
Another attempt to fix signedness warnings

djs55:
This is an experiment similar to #261 and #266.

Rather than cast everything to `(int)` I've tried to plumb though the type `enum vm_reg_name` and only cast to `(int)` at the lower level.


rn:
I prefer this approach, using `enum`s, over the simple casts to `(int)`. CI failure seems unrelated...

djs55:
@rn thanks for the feedback! I think I've finally managed to fix the CI: I've bumped the `Xcode` version the `opam` version for OCaml and worked around a couple of failures. Let me know what you think.

Issue #94, opened 2016-12-12
Deadlock in AHCI stop blockif_cancel due to lack of SIGCONT signal delivery


ijc25:
A long running blk io operation (we saw it with a long slow TRIM operation, but expect anything longer than ~30s would do) can cause a Linux guest to attempt to stop and reset the AHCI controller.

The issue appears to be that the `SIGCONT` which is sent by `blockif_cancel` and expected to be delivered via the mevent thread calling back into `blockif_sigcont_handler` never arrives and so `blockif_cancel` blocks forever. The interesting backtrace is:

```
  thread #13: tid = 0x2d000a5, 0x00007fff8c0aac8a libsystem_kernel.dylib`__psynch_cvwait + 10, name = 'vcpu:1'
    frame #0: 0x00007fff8c0aac8a libsystem_kernel.dylib`__psynch_cvwait + 10
    frame #1: 0x00007fff8c19496a libsystem_pthread.dylib`_pthread_cond_wait + 712
    frame #2: 0x000000010ea97316 com.docker.hyperkit`pci_ahci_write + 128 at block_if.c:899 [opt]
    frame #3: 0x000000010ea97296 com.docker.hyperkit`pci_ahci_write [inlined] ahci_port_stop(p=<unavailable>) + 180 at pci_ahci.c:445 [opt]
    frame #4: 0x000000010ea971e2 com.docker.hyperkit`pci_ahci_write [inlined] pci_ahci_port_write(sc=<unavailable>, offset=<unavailable>, value=<unavailable>) + 492 at pci_ahci.c:2051 [opt]
    frame #5: 0x000000010ea96ff6 com.docker.hyperkit`pci_ahci_write(vcpu=<unavailable>, pi=<unavailable>, baridx=<unavailable>, offset=<unavailable>, size=<unavailable>, value=<unavailable>) + 498 at pci_ahci.c:2156 [opt]
    frame #6: 0x000000010ea9c974 com.docker.hyperkit`pci_emul_mem_handler(vcpu=1, dir=<unavailable>, addr=<unavailable>, size=<unavailable>, val=<unavailable>, arg1=0x00007fd1ca7000a0, arg2=<unavailable>) + 284 at pci_emul.c:394 [opt]
[...]

PR #272, opened 2020-02-21
virtio-blk: implement DISCARD

djs55:
The protocol defines DISCARD in v1.1:

https://docs.oasis-open.org/virtio/virtio/v1.1/cs01/virtio-v1.1-cs01.pdf

In this simple implementation we advertise a maximum of 1 segment per request, which fits the backend API `block_delete(_, offset, length)`. Experimentally Linux 5.4.9 seems to send one request at a time (tested by deleting files and then executing `fstrim`)

Note we only advertise the feature if the backing file itself supports it.

Switching from AHCI `virtio-blk` works around the known deadlock in the AHCI implementation #94

### DIFF
--- a/src/lib/pci_virtio_block.c
+++ b/src/lib/pci_virtio_block.c
@@ -63,6 +63,8 @@
 #define	VTBLK_F_BLK_SIZE	(1 << 6)	/* cfg block size valid */
 #define	VTBLK_F_FLUSH		(1 << 9)	/* Cache flush support */
 #define	VTBLK_F_TOPOLOGY	(1 << 10)	/* Optimal I/O alignment */
+#define	VTBLK_F_DISCARD		(1 << 13)	/* Supports DISCARD / TRIM / WRITEZEROES */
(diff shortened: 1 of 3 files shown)


src/lib/pci_virtio_block.c
```diff
@@ -340,8 +396,16 @@ pci_vtblk_init(struct pci_devinst *pi, char *opts)
 
 	pthread_mutex_init(&sc->vsc_mtx, NULL);
 
+	/* Customise the capabilities per-device */
+	vi_consts = (struct virtio_consts*)malloc(sizeof(struct virtio_consts));
```
avsm: not testing for malloc failure here?

djs55: Ooh yes good point. I've added and force-pushed a NULL check on the next line ⬇️

avsm: Looks good. There is a calloc a few lines above also unchecked. But that’s unrelated to this PR

djs55:
Thanks!

PR #277, opened 2020-03-09
test: attempt to make CI more robust

djs55:
The CircleCI nested virtualisation often fails with

```
Kernel panic - not syncing: IO-APIC + timer doesn't work!  Boot with apic=debug and send a report.  Then try booting with the 'noapic' option.
```

Let's try booting the test kernels with the `noapic` option.


djs55:
OK this has now been green for 5 builds in a row. I'll merge this to fix CI on master.

PR #285, opened 2020-05-20
vmx: inherit defaults from hypervisor

djs55:
There have been some observed initialisation failures on new hypervisors / hardware, caused by us not recognising a new hypervisor feature. Previously the code would print an error and exit if it encountered an unknown feature.

However the hypervisor is able to provide a safe default, so we should use that instead. Setting the features we want is a VMCS write and querying the default fields is a VMCS read.

Therefore this patch
- moves the capability probing into the vCPU setup where a VMCS read is possible
- initialises the feature bits with a VMCS read (previously these were all 0)
- removes the failure case when we encounter an unknown feature bit (since the default will be safe)


samoht:
A bit hard to test but for a quick glance of that patch this LTGM

PR #286, opened 2020-06-06
ocaml: update dependencies and harden the notification loop

djs55:
- update to `opam` package manager version 2
- refresh all the dependencies

There is some evidence that the notification loop is spinning, due to an `Lwt_unix.read` call returning `-1`: https://github.com/docker/for-mac/issues/3499#issuecomment-623960890 . This is not supposed to happen but perhaps it is possible if the fd is set to the wrong blocking mode, such that `-1`, `EAGAIN` are returned but not handled correctly.

This PR 
- ensures the non-blocking mode is set correctly
- logs and exits if the call returns `<0`, which will highlight the problem and prevent spinning


djs55:
@samoht WDYT of this, in particular https://github.com/moby/hyperkit/pull/286/commits/6414a3c33383c011044ba1df8f6fd4965a52118b and https://github.com/moby/hyperkit/pull/286/commits/e88ddf6badb17c44b357630b5d6511e3511843c1 ?

samoht:
LGTM

PR #288, opened 2020-07-06
tty: write filename in a file, don't symlink to /dev

djs55:
If the osquery tool[1] is running and scanning a directory containing the `tty` symlink, it will follow the link to `/dev/ttys*` and start some kind of infinite loop involving the debug console. This causes hyperkit's CPU to spike and for the console itself to be spammed with it's own output.

The easiest workaround is to not create the symlink, but write the name of the tty in a file instead. Therefore for debugging we can still run

```
screen $(cat ~/Library/Containers/com.docker.docker/Data/vms/0/tty)
```
(or equivalent)

Reported originally around here[2]

[1] https://osquery.io/downloads/
[2] https://github.com/docker/for-mac/issues/3499#issuecomment-639140844


djs55:
Thanks!

Issue #298, opened 2020-10-15
Replace unix.SYS_IOCTL with unix.Ioctl{Get,Set}Termios


priyawadhwa:
I'm trying to import OpenTelemetry into my project, but Darwin builds are failing with this error:

```
# github.com/moby/hyperkit/go
../go/pkg/mod/github.com/moby/hyperkit@v0.0.0-20171020124204-a12cd7250bcd/go/pty_util_darwin.go:24:31: undefined: unix.SYS_IOCTL
```

The bug seems to be here:
https://github.com/moby/hyperkit/blob/b54460a1b41b487262d03c9685ae8b50dbc6607a/go/pty_util_darwin.go#L24

Go recently updated the `sys` package to remove direct system calls on Darwin (https://github.com/ory/dockertest/issues/212#issuecomment-685162574)

Other moby projects have merged PRs to update the system call (https://github.com/moby/term/pull/16). 

Would it be possible to update this syscall? I'm happy to open a PR if that helps.

PR #299, opened 2020-10-15
Use unix.IoctlSetTermios instead of ioctl system call on darwin

priyawadhwa:
This is necessary due to an upgrade in Golang's sys package, which no longer allows for direct system calls on darwin.

Fixes https://github.com/moby/hyperkit/issues/298

This code was copied from https://github.com/containerd/console/blob/2f74731477a5334b870605b261bcfb71ba34c56a/tc_unix.go, which a comment at the top of the file says the old code came from.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,5 +5,5 @@ go 1.14
 require (
 	github.com/mitchellh/go-ps v1.0.0
 	github.com/stretchr/testify v1.6.1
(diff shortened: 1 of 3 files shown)


priyawadhwa:
Hey @djs55 could you please take a look at this?

djs55 (approved):
LGTM!

PR #302, opened 2020-11-19
Fix for multiboot loader

agustingianni:
Hyperkit's multiboot loader incorrectly loads the firmware into guest memory.

This patch fixes the offset used to load the firmware image by using the offset of the first (smallest offset) loaded program segment in an elf32 firmware.

In essence what was happening before this patch was that hyperkit would load the entire ELF32 binary into memory from the start of the image. In some cases this works due to how some firmware image are built, but the correct behavior I think should be to find the offset instead of having to apply some link time tricks.

### DIFF
--- a/src/lib/firmware/multiboot.c
+++ b/src/lib/firmware/multiboot.c
@@ -222,7 +222,7 @@ multiboot_parse_elf(struct multiboot_state *s, struct image *img)
 {
 	struct elf_ehdr *ehdr = img->mapping;
 	struct elf_phdr *phdr;
-	uint32_t low = (uint32_t)-1, high = 0, memsize, addr, entry;
+	uint32_t low = (uint32_t)-1, high = 0, memsize, addr, entry, offset = (uint32_t)-1;
 	int i;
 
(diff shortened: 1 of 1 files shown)


djs55:
The CI failure looks unrelated:
```
undef: _bigstringaf_blit_to_bytes
undef: _bigstringaf_memcmp_bigstring
undef: _bigstringaf_memcmp_string
undef: _bigstringaf_blit_from_bytes
undef: _bigstringaf_blit_to_bigstring
Undefined symbols for architecture x86_64:
  "_bigstringaf_blit_to_bytes", referenced from:
      _camlBigstringaf__fun_1264 in mirage_block_ocaml.o
      _camlBigstringaf__substring_313 in mirage_block_ocaml.o
      _camlBigstringaf__to_string_320 in mirage_block_ocaml.o
      _camlBigstringaf__blit_to_bytes_407 in mirage_block_ocaml.o
      _camlBigstringaf__1 in mirage_block_ocaml.o
```

djs55 (approved):
LGTM

agustingianni:
Thank you very much for merging this!

PR #311, opened 2021-05-08
vmm: Hide host cpuid 0x15 TSC/Crystal ratio/freq info from guest

rn:
Further to the details below, with this fix booting a 5.10 kernel:
```
(ns: getty) linuxkit-025000000002:~# cat /proc/cpuinfo | grep bogo
bogomips        : 8000.00
bogomips        : 7999.99
bogomips        : 7999.89
bogomips        : 7986.02
```

Without:
```
(ns: getty) linuxkit-025000000002:~# cat /proc/cpuinfo | grep bogo
bogomips        : 7999.96
bogomips        : 1433.51
bogomips        : 1429.48
bogomips        : 1429.75
```

See also: https://github.com/linuxkit/linuxkit/pull/3636

This is a direct port of FreeBSD commit:
b4745ddd0511 ("Hide host CPUID 0x15 TSC/Crystal ratio/freq info from guest")
by grehan <grehan@FreeBSD.org>

Docker Desktop for Mac reverted Linux commit 2420a0b1798d ("x86/tsc: Set
LAPIC timer period to crystal clock frequency") as a workaround. This should
be the proper, ie don't report bogus crystal clock in cpuid and instead let
the kernel calibrate as it used to.

To quote the bhyve/FreeBSD commit message:

    Hide host CPUID 0x15 TSC/Crystal ratio/freq info from guest

    In recent Linux (5.3+) and OpenBSD (6.6+) kernels, and with hosts that
    support CPUID 0x15, the local APIC frequency is determined directly
    from the reported crystal clock to avoid calibration against the 8254
    timer.

    However, the local APIC frequency implemented by bhyve is 128MHz, where
    most h/w systems report frequencies around 25MHz. This shows up on
    OpenBSD guests as repeated keystrokes on the emulated PS2 keyboard
    when using VNC, since the kernel's timers are now much shorter.

    Fix by reporting all-zeroes for CPUID 0x15. This allows guests to fall
    back to using the 8254 to calibrate the local APIC frequency.

    Future work could be to compute values returned for 0x15 that would
    match the host TSC and bhyve local APIC frequency, though all dependencies
    on this would need to be examined (for example, Linux will start using
    0x16 for some hosts).

    PR:     246321
    Reported by:    Jason Tubnor (and tested)
    Reviewed by:    jhb
    Approved by:    jhb, bz (mentor)
    MFC after:      3 days
    Differential Revision:  https://reviews.freebsd.org/D24837

Signed-off-by: Rolf Neugebauer <rn@rneugeba.io>


fredericdalleau (approved):
LGTM, definitively an hypervisor fix!

PR #313, opened 2021-06-24
Various fixes

fredericdalleau:
This series add some fixes for various cases that have been reported that allow the guest to trigger a crash of the virtual machine.

The first two are NULL pointer dereference and are already handled in bhyve.
https://cgit.freebsd.org/src/tree/usr.sbin/bhyve/virtio.c#n606
https://cgit.freebsd.org/src/tree/usr.sbin/bhyve/virtio.c#n725

The other two are unhandled returned errors in vq_getchain. For now, trace and return as in pci_vt9p_notify.

Let me know what you think.


djs55 (approved):
LGTM! Thanks

PR #318, opened 2021-09-14
Fix assertion turned into assignment

fredericdalleau:
The assertion is supposed to check the value, but the typo makes it
an assignment. Hence it is always true.

Fixes GHSL-2021-058


djs55 (approved):
LGTM (oops!)

rn (approved):

PR #324, opened 2021-10-15
error: implicit conversion loses precision when running make (Apple clang version 13.0.0)

rheno:


rheno:
Signed-off-by: rheno <rhenobudiasa@yahoo.com>

